PR #214, opened 2026-02-09
refactor: rename internal functions to match public API

andymai:
## Summary

- Renames 29 internal function definitions across 9 `*Fns.ts` files to match their public API names (e.g. `fuseShape` → `fuse`, `translateShape` → `translate`, `filletShape` → `fillet`)
- Switches `topology/api.ts` and `operations/api.ts` to namespace imports (`import * as transforms`) to avoid name collisions with the locally-defined wrapper functions
- Preserves backward-compatible re-exports (`fromBREP as deserializeShape`, `chamferDistAngle as chamferDistAngleShape`)

## Motivation

AI agents and developers see different function names depending on whether they read the public API (`fuse`, `translate`, `fillet`) or the internal source files (`fuseShape`, `translateShape`, `filletShape`). This creates confusion about which names to use. After this change, internal and public names are consistent.

## Test plan

- [x] `npm run typecheck` — no type errors
- [x] `npm run test` — all 1628 tests pass
- [x] `npm run check:boundaries` — layer boundaries intact
- [x] `npm run lint` — 0 errors (50 pre-existing deprecation warnings)

### DIFF
--- a/llms.txt
+++ b/llms.txt
@@ -85,10 +85,10 @@ resolveDirection([a, b, c]): Vec3 // identity
 ### Shape Serialization
 
 ```typescript
-import { toBREP, deserializeShape } from 'brepjs';
+import { toBREP, fromBREP } from 'brepjs';
 
 const brep = toBREP(shape);              // shape → BREP string
-const restored = deserializeShape(brep); // BREP string → AnyShape
+const restored = fromBREP(brep);         // BREP string → AnyShape
 ```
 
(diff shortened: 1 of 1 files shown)


andymai:
@greptile-apps Please re-review — the two flagged issues (stale syntax highlight keywords and ambient type declarations) have been fixed in 5a0c19a. Also updated 4 additional site files that referenced old function names.

PR #783, opened 2026-04-13
fix(operations): fix pocket/boss positioning and extrude validation

andymai:
## Summary

- **pocket()/boss() not positioning profile on target face**: The profile wire was always at Z=0 — never translated to the target face center. A pocket on a box top face (Z=20) would create a tool from Z=0 to Z=-5, missing the box entirely. Now translates the profile to `faceCenter(targetFace)` before extruding.
- **extrude() exact zero-length check**: Used `=== 0` which misses near-degenerate vectors like `[1e-16, 0, 0]`. Now uses `< 1e-10` threshold consistent with other operations.
- **extrudeAll() silently passes zero-length entries**: The batch path didn't validate entry lengths, passing `length: 0` to the kernel. Now validates each entry upfront.
- **revolve() missing try/catch**: Kernel exceptions propagated uncaught, violating the `Result<T>` return contract. Now wraps in try/catch like `extrude()`.
- **mirrorJoin() zero-length normal**: No validation — a `[0,0,0]` normal would crash the kernel. Now validates with `vecIsZero()` like `drill()` does.

## Test plan

- [x] Pocket/boss tests strengthened to verify exact volume change (`toBeCloseTo`) instead of weak inequalities
- [x] All OCCT kernel tests pass with exact volume assertions
- [x] brepkit pocket/boss correctly skipped (face translate requires missing WASM exports)
- [x] Full test suite: same pass/fail count as main (29 failed / 101 failed — all pre-existing)
- [x] Typecheck, lint, boundary check, pattern check all clean

### DIFF
--- a/src/operations/compoundOpsFns.ts
+++ b/src/operations/compoundOpsFns.ts
@@ -173,13 +173,16 @@ export function pocket<T extends Shape3D>(shape: Shapeable<T>, options: PocketOp
   if (isErr(targetResult)) return targetResult as Result<T>;
   const targetFace = targetResult.value;
   const normal = normalAt(targetFace);
+  const center = faceCenter(targetFace);
   const w = toWire(profile);
 
   const faceResult = _makeFace(w);
   if (isErr(faceResult)) return faceResult as Result<T>;
 
+  // Position the profile on the target face before extruding
+  const positioned = translate(faceResult.value, center);
   const extDir = vecScale(vecNormalize(normal), -depth);
-  const toolResult = extrude(faceResult.value, extDir);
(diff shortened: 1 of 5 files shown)


andymai:
**Re: Greptile review — orphaned WASM face handle after translate**

Acknowledged. This follows the same pattern as other compound ops in this file — e.g. `drill()` orphans its `_makeCylinder` result after passing to `cut()`, and the old `pocket()` code orphaned `faceResult.value` after `extrude()` consumed it. None of the compound operations explicitly dispose intermediates; they rely on GC. A broader cleanup to add `DisposalScope` to all compound ops would be good but is a separate effort.

PR #790, opened 2026-04-21
feat(kernel): implement exportSTEP{Assembly,Configured} on occt-wasm

andymai:
## Summary

Second item on the occt-wasm parity-gap roadmap. Implements `exportSTEPAssembly` and `exportSTEPConfigured` by composing the existing `createXCAFDocument` + `writeXCAFToSTEP` primitives that were already implemented on occt-wasm. Removes two `notImplemented` stubs.

Previously, `tests/stepConfigFns.test.ts` had a lenient `isOk(result) ? ... : expect(error.code === 'STEP_EXPORT_CONFIGURED_FAILED')` branch to accept occt-wasm's stub failure. Those tests now take the happy path on occt-wasm.

## Known limitation

`options.unit`, `options.modelUnit`, and `options.schema` are currently ignored on occt-wasm — the underlying C++ facade (`occt-wasm@1.5.0`) exposes `writeXCAFToSTEP(docId)` without configuration parameters, and does not expose `Interface_Static` access. Output defaults to OCCT's `MM` + schema 5.

Full configurability requires an upstream binding change (e.g. `writeXCAFToSTEPConfigured(docId, unit, modelUnit, schema)`). Tracking as a follow-up.

## What's left in the "STEP assembly + config helpers" gap

The four adapter stubs `wrapString`, `wrapColor`, `configureStepUnits`, `configureStepWriter` remain as `notImplemented`. They are **OCCT-adapter-internal** helpers exposed by the `KernelIOOps` interface — they leak OCCT Embind types (`TCollection_ExtendedString_2`, `Interface_Static`, `STEPCAFControl_Writer`) into the kernel abstraction. Since occt-wasm now implements `exportSTEPConfigured` directly via composition, these helpers are never called on the occt-wasm path.

Cleaning them up is a broader kernel-interface refactor (remove from `KernelIOOps`) and out of scope for this PR.

## Test plan

- [x] `tests/stepConfigFns.test.ts` — all 9 tests pass on occt, occt-wasm, brepkit

### DIFF
--- a/src/kernel/occtWasm/occtWasmAdapter.ts
+++ b/src/kernel/occtWasm/occtWasmAdapter.ts
@@ -2108,7 +2108,11 @@ export class OcctWasmAdapter implements KernelAdapter {
   exportSTEPAssembly(parts: StepAssemblyPart[], _options?: { unit?: string }): string {
     if (parts.length === 0) return '';
     const doc = this.createXCAFDocument(parts);
-    return this.writeXCAFToSTEP(doc);
+    try {
+      return this.writeXCAFToSTEP(doc);
+    } finally {
+      doc.delete();
+    }
   }
 
   export3MF(_shape: KernelShape, _tolerance: number): ArrayBuffer {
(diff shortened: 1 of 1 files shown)


andymai:
Good catch — fixed in `10f2fb5`. Both methods now wrap `writeXCAFToSTEP` in `try/finally { doc.delete(); }` matching the reference pattern in `src/operations/exporterFns.ts:80-88`.

Note: on the current `occt-wasm@1.5.0` arena, `delete()` on the JS handle is a noop (the arena frees on kernel teardown, not per-handle). But honoring the interface contract (`ioOps.ts:50` — "Caller must delete the returned handle") keeps the code correct for any kernel and prevents real leaks once per-handle disposal lands.

PR #792, opened 2026-04-21
feat(kernel): implement exportOBJ on occt-wasm

andymai:
## Summary

Third roadmap PR for occt-wasm parity. Implements `exportOBJ(shape, tolerance): ArrayBuffer` by composing the existing `mesh()` tessellation primitive with a pure-TS Wavefront OBJ serializer.

**Goes beyond OCCT parity:** the OCCT default adapter throws `"exportOBJ is only available with the brepkit kernel"` (`src/kernel/occt/defaultAdapter.ts:1905`). occt-wasm now implements it natively, matching brepkit's functionality via the existing C++ tessellation facade — no upstream work needed.

## Format

Standard Wavefront OBJ:
- Vertices: `v x y z`
- Normals: `vn x y z`
- Faces: `f a//na b//nb c//nc` (1-based, triangle-only)

Angular deflection defaults to 0.5 rad (OCCT standard). Linear deflection comes from the `tolerance` parameter.

## Test plan

- [x] New smoke test in `tests/meshFns.test.ts` — asserts non-empty output + structural OBJ lines (`/^v /`, `/^f /`) — passes on occt, occt-wasm, and brepkit
- [x] OCCT branch correctly catches the pre-existing "only available with the brepkit kernel" throw

## Next in roadmap

- PR 2/3: `exportPLY` (same pattern, different format)
- PR 3/3: `exportGLB` (binary glTF — heavier)
- Cleanup PR: match OCCT throw messages for `import3MF`/`importGLB`/`importOBJ`/`export3MF` (features that remain brepkit-exclusive)

### DIFF
--- a/src/kernel/occtWasm/occtWasmAdapter.ts
+++ b/src/kernel/occtWasm/occtWasmAdapter.ts
@@ -2144,12 +2144,21 @@ export class OcctWasmAdapter implements KernelAdapter {
       const o = i * 3;
       lines.push(`vn ${n[o] ?? 0} ${n[o + 1] ?? 0} ${n[o + 2] ?? 0}`);
     }
-    const triCount = t.length / 3;
-    for (let i = 0; i < triCount; i++) {
-      const a = (t[i * 3] ?? 0) + 1;
-      const b = (t[i * 3 + 1] ?? 0) + 1;
-      const c = (t[i * 3 + 2] ?? 0) + 1;
+    const pushTri = (offset: number) => {
+      const a = (t[offset] ?? 0) + 1;
+      const b = (t[offset + 1] ?? 0) + 1;
+      const c = (t[offset + 2] ?? 0) + 1;
       lines.push(`f ${a}//${a} ${b}//${b} ${c}//${c}`);
+    };
+    if (result.faceGroups.length > 0) {
+      for (const group of result.faceGroups) {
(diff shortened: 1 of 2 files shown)


andymai:
Addressed two of the three P2 findings in `c3291f3`:

1. **Face-group markers** — `exportOBJ` now emits `g face_<hash>` when `result.faceGroups` is non-empty, matching the pattern in `src/io/objExportFns.ts:54-74`. Uses `faceHash` since that's what `KernelMeshResult.faceGroups` carries (the pure-TS `ShapeMesh` uses `faceId` — different layers, same semantics).
2. **`expect.hasAssertions()`** — added as a guard so the try/catch pattern can't pass vacuously.

Deferring the third finding (extract duplicated serialization to a Layer 0 util). The kernel/Layer-2 boundary prevents importing from `src/io/` as you noted — a shared util is the right long-term answer, but it's a cross-cutting refactor that should land as its own PR alongside the PLY/GLB work (same pattern will apply there).

Also: the `benchmark` check failed with a +82% regression on `gridfinity-bin-v2` — this benchmark exercises the bin-construction path (sketch extrude → shell → sweep lip → fillet) and doesn't touch `exportOBJ`, so the regression is almost certainly CI flake. The new commit will re-run the benchmark.

PR #796, opened 2026-04-21
feat(kernel): implement exportGLB on occt-wasm

andymai:
## Summary

Final mesh-format PR (3/3 in the beyond-OCCT-parity roadmap). Implements `exportGLB(shape, tolerance): ArrayBuffer` on occt-wasm by composing `this.mesh()` tessellation with a pure-TS binary glTF 2.0 writer.

**Goes beyond OCCT parity:** OCCT's `defaultAdapter` throws `"exportGLB is only available with the brepkit kernel"` (`src/kernel/occt/defaultAdapter.ts:1901`). occt-wasm now emits a valid GLB 2.0 container natively.

## Format

GLB 2.0 container:

```
Header (12 bytes):
  magic     : u32 LE = 0x46546C67 ('glTF')
  version   : u32 LE = 2
  length    : u32 LE = total bytes

Chunk 0 (JSON):
  chunkLength : u32 LE (padded to 4-byte boundary)
  chunkType   : u32 LE = 0x4E4F534A ('JSON')
  chunkData   : UTF-8 manifest, padded with ASCII space 0x20

Chunk 1 (BIN):
  chunkLength : u32 LE (padded to 4-byte boundary)
  chunkType   : u32 LE = 0x004E4942 ('BIN\0')
  chunkData   : positions (VEC3 float) | normals (VEC3 float) | indices (UINT32)
```

Manifest references three accessors (POSITION, NORMAL, indices) via three bufferViews over a single interleaved binary buffer. POSITION accessor includes required min/max bounds. Indices use `UNSIGNED_INT` (componentType 5125) to avoid overflow on large meshes.

## Endianness

`DataView` is used for all multi-byte header writes to guarantee little-endian order per GLB spec. Typed-array bodies (Float32 positions/normals, Uint32 indices) are copied via `Uint8Array` views — valid on all JS runtimes in practice (all little-endian).

## Test plan

- [x] New smoke test in `tests/meshFns.test.ts` — asserts GLB magic bytes (`0x46546C67`), version 2, and `length` field matches `ArrayBuffer.byteLength`
- [x] OCCT branch correctly catches the "only available with the brepkit kernel" throw
- [x] `npx tsx scripts/check-patterns.ts` — bounds + manifest construction extracted to module-level helpers to satisfy max-function-lines

## Roadmap position

- [x] PR #788: cone/torus direction fix
- [x] PR #790: exportSTEP{Assembly,Configured}
- [x] PR #792: exportOBJ
- [x] PR #794: exportPLY
- [x] **This PR: exportGLB**
- [ ] Final: cleanup PR to match OCCT throw messages for remaining import stubs (`import3MF`/`importGLB`/`importOBJ`/`export3MF`) — features that remain brepkit-exclusive

### DIFF
--- a/src/kernel/occtWasm/occtWasmAdapter.ts
+++ b/src/kernel/occtWasm/occtWasmAdapter.ts
@@ -251,6 +251,7 @@ function computePositionBounds(positions: Float32Array, vCount: number): Vec3Bou
 
 function buildGltfManifest(
   vCount: number,
+  nCount: number,
   iCount: number,
   posBytes: number,
   nrmBytes: number,
@@ -288,7 +289,7 @@ function buildGltfManifest(
         min: bounds.min,
         max: bounds.max,
       },
-      { bufferView: 1, componentType: 5126, count: vCount, type: 'VEC3' },
+      { bufferView: 1, componentType: 5126, count: nCount, type: 'VEC3' },
       { bufferView: 2, componentType: 5125, count: iCount, type: 'SCALAR' },
     ],
   };
@@ -2206,6 +2207,7 @@ export class OcctWasmAdapter implements KernelAdapter {
     const normals = result.normals;
     const indices = result.triangles;
     const vCount = positions.length / 3;
+    const nCount = normals.length / 3;
     const iCount = indices.length;
 
     // Binary buffer: positions | normals | indices. All components are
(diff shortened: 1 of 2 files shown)


andymai:
Addressed the P2 in `15c12e5` — NORMAL accessor now uses `nCount` (derived from `normals.length / 3`) rather than `vCount`. Currently equal since OCCT tessellation returns per-vertex normals, but binding each accessor to its own data makes the invariant explicit.

PR #801, opened 2026-04-21
fix(kernel): real per-face evolution in occt-wasm applyComposedTransformWithHistory

andymai:
## Summary

Third PR in the Group A sub-roadmap. The existing `applyComposedTransformWithHistory` on occt-wasm only handled `transformHandle.__type === 'transform_matrix'` and, even in that branch, synthesized a semantically broken evolution where every input face hash mapped to the output **shape** hash (not a face hash). Downstream `propagateAllMetadata` then silently failed to propagate origins/tags/colors across the transform on occt-wasm.

This PR both closes the `notImplemented` fallback and fixes the evolution quality.

## Changes

1. **Handle dispatch** — delegate to `this.transform()` which already accepts plain `number[]`, `{ matrix: [] }`, `{ elements: [] }`, and the `__type`-branded variant. Matrix-like handles all work; `notImplemented` fallback removed.

2. **Real per-face evolution** — pair input/output face hashes by iteration index. Rigid and affine transforms preserve topology, and both the caller (`collectInputFaceHashes` in `metadataPropagation.ts`) and this method iterate via `kernel.iterShapes(..., 'face')` — order-stable, so index-correspondence holds.

3. True C++-tracked per-face evolution for generic composed transforms would require a new facade method — `translateWithHistory` and `rotateWithHistory` exist, but not for composed matrix transforms. Best-effort TS synthesis is correct under the topology-preservation invariant that composed transforms satisfy.

## Why this matters

`propagateAllMetadata` reads `evolution.modified` to move face origins, tags, and colors from input to output faces. With the previous (broken) synthesis, every input face hash was mapped to a value that doesn't appear in the result's face-hash space, so metadata lookups silently found nothing. occt-wasm users relying on `transformCopy` would have had metadata silently lost. Now it works.

## Test plan

- [x] Existing `transformCopy` tests (7 across 3 kernels) still pass
- [x] New regression test in `tests/transformCopy.test.ts` — asserts evolution has one entry per input face and every mapped output hash is a real face hash in the result. Passes on occt, occt-wasm, and brepkit
- [x] `npm run typecheck` clean, `npm run lint` clean

## Roadmap

- [x] PR #788: cone/torus direction
- [x] PRs #790/#792/#794/#796/#797: STEP + mesh formats
- [x] PR #798: 2D primitives
- [x] PR #800: iterShapeList informative throw
- [x] **This PR:** applyComposedTransformWithHistory
- [ ] Group A final: `helicalSweep` (compose helix wire + sweep)

### DIFF
--- a/src/kernel/occtWasm/occtWasmAdapter.ts
+++ b/src/kernel/occtWasm/occtWasmAdapter.ts
@@ -1994,22 +1994,31 @@ export class OcctWasmAdapter implements KernelAdapter {
     inputFaceHashes: number[],
     hashUpperBound: number
   ): OperationResult {
-    if (transformHandle.__type === 'transform_matrix') {
-      // Apply the matrix as a generalTransform, then synthesize evolution
-      const result = this.transform(shape, transformHandle);
-      const modified = new Map<number, number[]>();
-      for (const h of inputFaceHashes) {
-        const resultHash = this.hashCode(result, hashUpperBound);
-        modified.set(h, [resultHash]);
+    // transform() dispatches across all matrix-like handle shapes:
+    // { __type: 'transform_matrix', matrix }, { matrix: [] }, { elements: [] },
+    // and plain number[]. It throws on anything unrecognizable.
+    const result = this.transform(shape, transformHandle);
+
+    // Synthesize per-face evolution by pairing input and output face hashes
+    // by iteration index. Affine transforms preserve topology, and both
+    // the caller (collectInputFaceHashes) and we use iterShapes(..., 'face')
+    // which is order-stable — the correspondence holds. True C++-tracked
+    // evolution would need a facade method for generic transforms (only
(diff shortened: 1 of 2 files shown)


andymai:
Deferring both P2 findings — both are speculative hardening with no current caller:

1. **Silent identity fallback** (Issue 1): lives inside `transform()`'s dispatch logic, not this method. Addressing it means changing `transform()`'s behavior, which is cross-cutting — it affects every caller, not just the one I touched. That's a separate refactor PR.

2. **Face-count mismatch warning** (Issue 2): `console.warn` would be the first console call in either adapter (`grep -c "console\." src/kernel/*Adapter.ts` → 0 in both). Adding it here sets a new pattern for the layer. Since no current path triggers the mismatch, I'd rather not introduce the pattern preemptively.

Both are legitimate observations worth tracking for a future hardening pass.

PR #827, opened 2026-05-06
feat(site)!: serve docs-site at root, playground at /playground

andymai:
## Summary

- Replace the React landing page with the VitePress docs site (#824) as the canonical homepage. The site/ app is stripped to the playground SPA and mounted at `/playground/` via a composite Vercel build that copies `site/dist/` into `docs-site/.vitepress/dist/playground/`.
- Move the share URL format from `#code/<lz>` to `?code=<lz>`. Legacy hash links are decoded on first load and silently rewritten to the new query form, so existing shared URLs keep working.
- Scope `COEP`/`COOP`/`CSP` headers to `/playground/*` only — VitePress pages no longer ship `require-corp`, so docs can use third-party fonts/assets without breakage.
- Port favicon set + manifest + theme color from `site/public/` to `docs-site/public/` so the docs head and the playground match visually.

### Why this PR

Two follow-ups:
1. The docs site (PR #824) is the more complete and accurate surface for users; the React landing page had drifted and was less useful.
2. The combined deployment lets us link to ` /playground` from any docs page without a cross-origin trip.

### What's deleted

- `site/src/components/landing/` (12 files)
- Gallery precompute script + `lib/meshLoader.ts` + `hooks/usePrecomputedGalleryMeshes.ts`
- `site/public/gallery-meshes/` (gitignored content)
- `react-router-dom` dependency

## Test plan

- [x] `npm run typecheck` (root and site)
- [x] `npm run build --workspace=site` produces a SPA with all asset URLs prefixed by `/playground/`
- [x] `cd docs-site && npm run build` produces VitePress dist
- [x] Composite copy: `mkdir -p docs-site/.vitepress/dist/playground && cp -r site/dist/. docs-site/.vitepress/dist/playground/` populates the merged tree
- [x] Static smoke test (`python3 -m http.server` over the merged dist):
  - `GET /` → 200 (docs home)
  - `GET /playground/` → 200 (playground SPA)
  - `GET /playground/wasm/brepjs_single.js` → 200
  - `GET /playground/index.html` → 200
- [x] Pre-commit hooks (lint, typecheck, boundaries, changed-file tests) pass
- [ ] Vercel preview build succeeds and routes both surfaces correctly
- [ ] Verify legacy hash share URLs (e.g. `/playground/#code/<lz>`) silently migrate to `/playground?code=<lz>` on first load

## Follow-up

A second PR will do a static-review pass over the playground UX for the bug categories called out separately (worker/state races, Monaco issues, viewer/3D, error handling and loading).

## Notes

- `docs-site/vercel.json` is left in place for an optional standalone deployment, but the root `vercel.json` is now the canonical project.
- `vitest.config.ts` excludes `docs-site/**` so test discovery does not pick up tests inside `docs-site/node_modules`.


andymai:
Addressed all three Greptile comments in 44cbdaa:

- **vercel.json (P1)**: Header source is now `/playground(/.*)?` — matches the bare `/playground` URL Vercel serves under cleanUrls + trailingSlash:false, while still being narrow enough not to false-match `/playgrounds` or `/playground-foo`. COEP/COOP/CSP now reach the playground entry point.
- **docs-site/.vitepress/config.ts (P2)**: Explicit `rel: 'noopener noreferrer'` on the playground nav item. Same fix on the home-page action in `docs-site/index.md`.
- **site/src/lib/urlCodec.ts (P2)**: Added a `console.warn` on decompress failure for both `?code=` and legacy `#code/` paths. The user-facing toast for this case lands in #828 (which depends on the toast store).

PR #854, opened 2026-05-07
fix(playground): evaluate user code as ES modules so docs links work

andymai:
## Summary

- The docs-site "Open in Playground" buttons URL-encoded `import { … } from 'brepjs/quick'` snippets, but the worker evaluated them with `new Function(code)` — a function body, not a module — so every shared link threw a syntax error before running. Even after stripping imports, docs blocks never returned anything, so the viewer stayed empty.
- Switched the worker to dynamic-import-of-blob-URL ESM evaluation: at init it builds a wrapper Blob that re-exports the loaded brepjs runtime via `self.__brepjs`; per-eval it rewrites `from 'brepjs'` / `from 'brepjs/quick'` specifiers in user code to that wrapper URL, then `import()`s the user blob. Render contract is now `export default <shape>;`.
- Cancellation is now per-eval-id (Set-based) so concurrent in-flight evals don't clobber each other when the user spams Run — the previous global flag had a race where eval#2 would clear eval#1's cancel state.
- Monaco's ambient `.d.ts` is regex-transformed into `declare module 'brepjs' { … }` / `declare module 'brepjs/quick' { … }` so editor imports resolve cleanly.
- Examples removed entirely (data, picker UI, screenshot/test scripts, `?example=` URL branch). Default snippet is a parametric stud-and-tube brick exercising primitives, iteration, fuse, `ofCurveType` finder, and chamfer.
- Tightened docs decorator filter to require both `from 'brepjs…'` and `^export default` in a block; added `export default <name>;` to ~35 runnable code blocks across `getting-started/`, `tasks/`, `migration/`, and `index.md`. Reference-only blocks intentionally have no button.

## Test plan

- [x] `npm run typecheck` (root + `site/`) passes
- [x] `npm run check:boundaries` passes
- [x] `npm run build` (root + `site/`) passes
- [x] `vitepress build docs-site/` passes
- [x] Headless probe: default stud-brick code renders to a 399×491 canvas
- [x] Headless probe: `?code=`-encoded ESM snippet (`import` + `export default`) decodes, runs through the new worker path, and produces a mesh
- [x] Headless probe: `first-solid.md` renders 4 working "Open in Playground" buttons with valid encoded URLs
- [ ] Verify cancellation race fix manually by spam-clicking Run with a slow snippet
- [ ] Spot-check a few playground links from each docs section open and run


andymai:
Thanks for the review. Both addressed in 2b5b6d3:

- **Issues 1 & 2** (partial-init window): widened the re-init guard to `brepjs && brepjsBlobUrl`. If `initFromOC` throws between the import and the wrapper-URL build, a subsequent init dispatch will now retry instead of falsely reporting success.
- **Issue 3** (`declare type` skipped): mirrored the optional `(?:declare\s+)?` from the keyword-declaration branch so both `type Foo =` and `declare type Foo =` are promoted to `export type Foo =`. The current auto-generated ambient file only emits the bare form, so this is a latent fix.

PR #856, opened 2026-05-07
feat(playground): click a face or edge to copy its finder predicate

andymai:
## Summary

The viewer is now interactive. Hover or click any face / edge in the rendered shape and the playground copies the tightest brepjs finder predicate that identifies it onto the clipboard, while the status bar shows a one-liner of its identity. This is the cross-cutting feature from the brepjs-studio cherry-pick path: serves first-timers (learn the API by clicking), docs visitors (verify a snippet's geometry matches expectation), and prototypers (skip the "how do I select that edge?" debug loop).

- **Worker (single-shape evals only):** switch face mesh to `toGroupedBufferGeometryData`, pass through `EdgeMesh.edgeGroups`, extract per-face metadata via `getFaces` + `getSurfaceType` + `measureArea` + `normalAt` keyed by `getHashCode`, same for edges via `getEdges` + `getCurveType` + `curveLength`.
- **`ShapeRenderer`:** registers per-face groups on the BufferGeometry; click reads `event.face.materialIndex`, looks up the matching `FaceInfo`, copies snippet, fires toast.
- **`EdgeRenderer`:** binary-search by vertex index over edge groups; bumped line-pick threshold so picking isn't pixel-perfect on hi-DPI screens.
- **Snippet shape:** partial predicate without `.findAll(...)` so it pastes into either the functional API (`<PASTE>.findAll(part)`) or the fluent wrapper (`shape(p).fillet(<PASTE>, 1.5)`). Faces emit `faceFinder().ofSurfaceType(T).inDirection(D).withArea({…})`; edges emit `edgeFinder().ofCurveType(T).withLength({…})`.
- **Selection** survives in `playgroundStore.selection`, cleared on every new render. Status bar formatters in `lib/selectionLabels.ts` borrowed from `brepjs-studio` and reconciled to brepjs's actual `SurfaceType` / `CurveType` enum names.

Multi-shape evals (`export default [a, b, c]`) skip the inspection metadata to keep the picking semantics unambiguous — this matches the playground's "deliberately lightweight" positioning vs. studio.

## Test plan

- [x] `npm run typecheck` (root + `site/`) passes
- [x] `npm run check:boundaries` passes
- [x] `npm run build` (site) passes
- [x] Headless probe: clicking the centre of the default stud-brick canvas fires the worker → renderer → store pipeline; status bar updates and toast confirms `Edge finder copied`
- [ ] Verify face click on a non-brick shape (e.g. a sphere) emits a `'SPHERE'` finder
- [ ] Verify the snippet pastes correctly into a functional-API call and produces the expected match


andymai:
Fixed in ae0bd9f5 — added a cleanup useEffect in both renderers that unconditionally resets `document.body.style.cursor = ''` on unmount. Closes the leak when a renderer disappears mid-hover (new eval drops the previous mesh, multi-shape switch flips `pickable` to false, etc.).

PR #875, opened 2026-05-07
feat(playground): first-run onboarding hint

andymai:
## Summary
Small dismissible pill at the bottom-center of the viewport: **"💡 Click a face or edge to copy a finder predicate"**. Closes the discoverability gap for first-time visitors who land on a docs link or shared playground URL — they previously had no signal that the 3D scene was interactive at all.

Auto-dismisses on the user's first selection (they clearly already discovered the flow) or on the close button. Persisted in \`localStorage\` so the pill never reappears once dismissed.

## Test plan
- [ ] Fresh tab in incognito → pill appears below the model
- [ ] Click a face → pill disappears, doesn't return on reload
- [ ] Fresh tab → pill appears, click X → pill disappears, doesn't return on reload
- [ ] Returning user (already dismissed) → pill stays hidden

### DIFF
--- a/site/src/components/playground/OnboardingHint.tsx
+++ b/site/src/components/playground/OnboardingHint.tsx
@@ -1,0 +1,72 @@
+import { useEffect, useState } from 'react';
+import { usePlaygroundStore } from '../../stores/playgroundStore';
+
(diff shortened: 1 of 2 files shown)


andymai:
@greptileai

PR #881, opened 2026-05-07
feat(playground): right-click menu for finder copy + reliable hover

andymai:
## Summary

- **Right-click menu**: replaces left-click-to-copy with a context menu that surfaces a "Copy finder predicate" action; left-click now only selects.
- **Hover fix**: every face/edge of a shape is now hoverable. Earlier, large +Z top faces of a stud brick were the only reliable hover targets.

## Hover root cause

Two compounding bugs in `ShapeRenderer.tsx` / `EdgeRenderer.tsx`:

1. `pointerOut` unconditionally nulled global `hoverEntity`. R3F can fire one renderer's `pointerOut` after a sibling's `pointerMove` already set hover state in the same tick, blanking valid hovers. The renderer now compares its own last-published id (`useRef`) against `getState().hoverEntity` and only nulls if it is still the active target.
2. `PICK_THRESHOLD_WORLD = 0.5` for line raycasting was huge relative to a 9.6mm stud-brick body; every side face partially fell inside line pick volumes, so edges always won the closest-hit race via `stopPropagation`. Tightened to `0.15`.

## Right-click menu

- New `ContextMenu.tsx`, viewport-clamped, modeled after `SelectionTooltip.tsx`.
- Dismisses on outside click, Escape, scroll, resize, camera drag, and new eval.
- `onContextMenu` on face mesh and edge `lineSegments` resolves the target, calls `event.nativeEvent.preventDefault()`, and opens the menu without mutating `selections`.
- Wrapper `<div>` on `ViewerPanel` suppresses the browser context menu on empty canvas. `onPointerMissed` distinguishes button=2 (close menu only) from left-click (close + clear selection).
- Toast still confirms clipboard write/fail; click handlers and copy logic are now decoupled.

## Messaging

- `OnboardingHint`, `SelectionTooltip` snippet preview, and `ShortcutHelp` updated to point users at right-click for the copy action.

## Test plan

- [ ] Hover every face of `studBrick(2, 4, 3)` (default code) — top, sides, stud sides, stud tops, bottom, cavity walls. Tooltip should follow the cursor on each.
- [ ] Hover every visible edge — should switch to edge tooltip, not get blocked by face hover.
- [ ] Left-click a face: selection commits, no clipboard write, no toast.
- [ ] Right-click a face: menu opens at cursor, "Copy finder predicate" copies the snippet, toast confirms.
- [ ] Right-click while menu is open elsewhere: replaces the menu at the new location.
- [ ] Escape, scroll, camera-drag, click-outside all dismiss the menu.
- [ ] Right-click empty canvas: no menu, no native browser menu.
- [ ] Right-click does not change `selections` even when right-clicking an unselected face.

### DIFF
--- a/site/src/components/playground/ContextMenu.tsx
+++ b/site/src/components/playground/ContextMenu.tsx
@@ -1,0 +1,110 @@
+import { useEffect, useLayoutEffect, useRef, useState } from 'react';
+import { usePlaygroundStore } from '../../stores/playgroundStore';
+import { useToastStore } from '../../stores/toastStore';
+import { buildEdgeFinderSnippet, buildFaceFinderSnippet } from '../../lib/finderSnippet';
+import { copyToClipboard } from '../../lib/copyToClipboard';
+import type { Selection } from '../../stores/playgroundStore';
+
+interface Props {
+  containerRef: React.RefObject<HTMLDivElement | null>;
+}
+
+const OFFSET_X = 4;
+const OFFSET_Y = 4;
+const EDGE_PADDING = 8;
+
+export default function ContextMenu({ containerRef }: Props) {
+  const contextMenu = usePlaygroundStore((s) => s.contextMenu);
+  const closeContextMenu = usePlaygroundStore((s) => s.closeContextMenu);
+  const addToast = useToastStore((s) => s.addToast);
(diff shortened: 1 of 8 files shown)


andymai:
@greptileai review

andymai:
@greptile-apps review

Issue #1026, opened 2026-05-19
Transitive dep advisories blocked by upstream pinning


andymai:
Tracking issue for dev-only dependency vulnerability alerts that cannot be cleared via `npm overrides` today.

## Blocked alerts

All 10 alerts surfaced by Dependabot affect dev/build dependencies of `apps/docs` and `apps/playground`. None reach the published `brepjs` runtime surface.

### dompurify (8 advisories) — blocked by `monaco-editor`

- `monaco-editor@0.55.1` declares `dompurify: 3.2.7` as an **exact** version pin (not a range).
- Forcing `dompurify: ^3.4.5` via `npm overrides` creates a resolution conflict; npm silently omits dompurify, which breaks the docs build.
- Dependabot matches alerts on `(package, vulnerable_version_range)` against every lockfile entry, so even though the mermaid nested copy is now on 3.4.5 (see below), all 8 advisories remain open because monaco-editor's 3.2.7 still satisfies every vulnerable range.
- **Unblock condition:** monaco-editor publishes a stable release that updates the pin (currently only `0.56.0-dev-*` exists). Track [microsoft/monaco-editor](https://github.com/microsoft/monaco-editor) releases.

### ~~dompurify in mermaid — intermittent override behavior~~ Resolved in #1043 (2026-05-20)

- Root cause was **lockfile inertia**, not a resolver bug. The nested `mermaid/dompurify` entry was written by Dependabot's mermaid 11.14.0 → 11.15.0 bump on 2026-05-11, when 3.4.2 was the latest version satisfying `^3.3.1`. dompurify subsequently shipped 3.4.3 (5/13), 3.4.4 (5/17, deprecated), and 3.4.5 (5/18). Nothing forces re-resolution of a transitive-of-a-transitive, so the lockfile stayed pinned.
- Verified: a clean install in `/tmp` today picks 3.4.5; deleting the brepjs lockfile and regenerating also picks 3.4.5 (and flips the hoist topology). The earlier "fresh lockfile still produces 3.4.2" observation was either an `rm -rf node_modules` (which doesn't touch the lockfile) or a registry warm-up artifact — 3.4.5 was only ~24h old when this issue was filed.
- Fix in #1043 was a surgical 3-field hand-edit of the lockfile (version + resolved + integrity). Avoided `npm update dompurify` because npm 11.6.1 strips `@emnapi/core` + `@emnapi/runtime` from the lockfile (separate known issue).
[...]

PR #1043, opened 2026-05-20
fix(deps): bump nested mermaid → dompurify 3.4.2 to 3.4.5

andymai:
## Summary
- Surgical lockfile bump: `node_modules/mermaid/node_modules/dompurify` from 3.4.2 → 3.4.5
- Chases the resolution "mystery" called out in #1026

## Why was it stuck at 3.4.2?
The lockfile entry was written by Dependabot's mermaid 11.14.0 → 11.15.0 bump on **2026-05-11**. At that moment, the latest dompurify satisfying mermaid's `^3.3.1` range was 3.4.2 — 3.4.3 didn't ship until 2026-05-13, 3.4.5 not until 2026-05-18. Since dompurify is a transitive of a transitive, nothing has forced npm to re-resolve it; the lockfile is just sticky.

Verified end-to-end:
- Standalone mermaid in /tmp picks 3.4.5 today
- Deleting the brepjs lockfile and regenerating picks 3.4.5 (and flips the hoist topology — root becomes 3.4.5, monaco-editor's 3.2.7 gets nested)
- Hand-edited current lockfile (this PR's diff) installs cleanly via `npm ci` in /tmp scratch

3.4.4 is deprecated ("Fixed a security issue introduced in 3.4.4"). 3.4.5 is the current `latest` dist-tag.

## Why a hand-edit instead of `npm update dompurify`?
Local npm 11.6.1 strips `@emnapi/core` + `@emnapi/runtime` from the top-level lockfile (known issue tracked in memory; CI rejects this). The hand-edit avoids that side-effect entirely — only the three lines for the nested dompurify change.

## Impact on #1026
Clears the 4 dompurify advisories on the mermaid path. The other 4 dompurify advisories (monaco-editor's exact-pinned 3.2.7) are still blocked upstream until monaco-editor ships a stable 0.56+.

## Test plan
- [ ] CI `npm ci` succeeds
- [ ] Docs build still renders mermaid diagrams
- [ ] Dependabot alerts on mermaid path drop from 10 → ~6

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8466,9 +8466,9 @@
       }
     },
     "node_modules/mermaid/node_modules/dompurify": {
-      "version": "3.4.2",
-      "resolved": "https://registry.npmjs.org/dompurify/-/dompurify-3.4.2.tgz",
-      "integrity": "sha512-lHeS9SA/IKeIFFyYciHBr2n0v1VMPlSj843HdLOwjb2OxNwdq9Xykxqhk+FE42MzAdHvInbAolSE4mhahPpjXA==",
+      "version": "3.4.5",
+      "resolved": "https://registry.npmjs.org/dompurify/-/dompurify-3.4.5.tgz",
+      "integrity": "sha512-OrwIBKsdNSVEeubdJ1HBv/wNENRM9ytAVCv7YXt//A3vPdVMNuACRqK9mXCGCBW2ln7BT/A4X0jXHo2Gu89miA==",
       "dev": true,
       "license": "(MPL-2.0 OR Apache-2.0)",
       "optionalDependencies": {


andymai:
Correction to the merge message: this PR clears **0 Dependabot alerts**, not 4.

Dependabot matches alerts to `(package, vulnerable_version_range)` against the lockfile as a whole, not to specific `node_modules` paths. Monaco-editor's still-pinned `dompurify@3.2.7` satisfies every vulnerable range covered by the 8 dompurify advisories, so they all remain open until #1026 is resolved upstream.

The bump is still a real security improvement — mermaid's runtime path no longer loads a vulnerable dompurify — but the Dependabot dashboard count is unchanged.

PR #1094, opened 2026-05-29
test(brepkit): un-skip cannedSketches face-offset divergence

andymai:
The brepkit `sketchFaceOffset` inward path now produces the expected shrunk area (~252.6 for a 20×20 box face offset by -2, well under the 400 threshold) instead of expanding to area 576.

This removes the stale `cannedSketches.faceOffset` skip entry so the test runs under the brepkit kernel.

Validated with a locally-built brepkit-wasm carrying the offset sign fix: the cross-kernel test `sketchFaceOffset shrinks a face inward` passes for brepkit, occt, and occt-wasm.

### DIFF
--- a/tests/helpers/kernelDivergences.ts
+++ b/tests/helpers/kernelDivergences.ts
@@ -96,14 +96,6 @@ export const divergences: DivergenceMap = {
     },
 
     // -----------------------------------------------------------------------
-    // cannedSketches.test.ts
-    // -----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


andymai:
Superseded by #1096, which un-skips the same cannedSketches.faceOffset divergence AND includes the required brepkit-wasm bump (2.101.2, contains offset fix #741) plus adapter wiring. This PR only removes the divergence entry without the bump, so its un-skipped test would fail against the pinned wasm. Closing as duplicate.

PR #1103, opened 2026-05-29
fix(playground): curved, pitched impeller blades for the axial fan example

andymai:
## Summary

Reworks the **axial cooling fan** example's impeller blades. The old blades were
flat radial paddles — sparse, un-angled, and nothing like a real fan rotor.

## What changed

- Each blade is now an **organically curved, pitched vane**: a rectangular
  cross-section is marched along a curling tangential path from hub to rim,
  twisting the pitch from steep at the root to shallow at the tip. A chain of
  per-segment convex hulls is fused so the blade scoops concavely the way a real
  impeller does — something a single straight hull can't express.
- Blade tips sit ~1 mm **inside** the air bore (was clipping through the frame
  wall before); 9 blades overlap into a solid-looking rotor.
- Thumbnail regenerated (`npm run thumbs`).

## Verification
- 174 tests pass (example eval+mesh regression + thumbnail-coverage gate).
- Playground typecheck clean; pre-push gate (`test:full` + `knip`) passed.

Before/after is visible in the regenerated `axial-fan.webp` thumbnail.


andymai:
Both addressed in the latest commit — `section(t: number)` annotation added and the `rOuter` comment reworded to reference the bore radius. No geometry change.

Issue #1102, opened 2026-05-29
Flaky CI: tests/2d.test.ts > CompoundBlueprint > 'basic properties and SVG' times out (~90s limit)


andymai:
## Summary
`tests/2d.test.ts` → `CompoundBlueprint > basic properties and SVG` (around `tests/2d.test.ts:320`) intermittently **times out at the 90000ms per-test limit** in the `test:full` run. It is not a logic failure — it occasionally runs just over the timeout (observed **91398ms** against the 90000ms limit), so it reds CI nondeterministically.

## Impact
This sits in the unconditional `test:full` path, so it intermittently fails CI on **unrelated** PRs and forces a manual job re-run. Observed reds on at least:
- PR #1080 (mesh angularTolerance bump) — failed here, passed on re-run
- PR #1096 (face-offset un-skip) — failed here, passed on re-run

In both cases the PR's own tests passed; only this property test timed out.

## Likely cause
It appears to be a `fast-check` property test whose per-case work (CompoundBlueprint property computation + SVG generation) is heavy enough that, under CI load or an unlucky seed/sample, total runtime crosses the 90s ceiling. It's a runtime/timeout issue, not a correctness one.

## Suggested fixes (any one)
- Reduce `fc` sample size for this test (`numRuns`) or shrink the generated input bounds.
- Raise the per-test timeout for this specific test (e.g. `it(..., { timeout: 180000 })`) if the coverage is intentionally large.
- Profile/optimize the CompoundBlueprint `basic properties and SVG` path (SVG serialization is a common hot spot) so a single run stays well under the limit.
- Optionally pin the `fast-check` seed to make runtime deterministic.

## Repro
Run the `test:full` suite (occt project); the timeout surfaces intermittently. Re-running the failed job currently clears it.

andymai:
Recurred on PR #1100 (origin-propagation fix, unrelated diff — touches only `originTrackingFns.ts` + a new test). The `test` job timed out at the 90000ms limit on this same case (`tests/2d.test.ts:320`); the PR's own tests passed. Confirmed locally that the test passes well under the limit (~37s for the whole file incl. WASM init). Cleared by re-running the failed job.

Adds to the impact list: #1080, #1096, #1100.

PR #1115, opened 2026-05-30
ci(hooks): align pre-push with the CI test gate (drop coverage)

andymai:
## Summary

Aligns the `pre-push` hook with what CI actually gates on, and gives a local lever to use more cores.

Pre-push ran `test:full` — the full suite **with v8 coverage**. But CI gates PRs on the sharded `test` job, which runs `test:ci` (full suite, **no coverage**); the `coverage` job runs `test:full` **main-only** and is deliberately excluded from `ci-pass` (informational, non-blocking). So the hook was *stricter than CI* and paid coverage instrumentation + lcov/text report generation on every push, for a gate CI doesn't enforce.

## Changes

- **`.husky/pre-push`** now runs `npm run test:ci` (full suite, no coverage) instead of `npm run test:full`. `knip` still runs.
- **`vitest.config.ts`** `maxWorkers` is now overridable via `VITEST_MAX_WORKERS`, **default unchanged at 4**. Invalid/empty values fall back to 4.
- Docs (`CLAUDE.md` hooks section) updated to match.

## Why keep the default at 4

CI's sharded `test` job runs `test:ci` with the config default on 16 GB runners. OCCT WASM linear memory grows monotonically across a fork's files, so more forks over-commit into swap and trip the per-test timeout — exactly the failure #1102 fixed by sharding. Raising the default would reintroduce that. The env override lets a high-RAM, many-core dev machine opt into more forks locally (e.g. `VITEST_MAX_WORKERS=8`) without touching CI. The `coverage` job is unaffected — it passes `--maxWorkers=2` on the CLI, which overrides the config regardless of the env var.

## Verification

- `maxWorkers` resolution: unset → 4, `VITEST_MAX_WORKERS=8` → 8, garbage → 4.
- `typecheck`, `prettier --check`, and `vitest list` all clean.
- This PR's own push dogfoods the new hook (runs `test:ci`).

No library code changes — `ci`-typed so it doesn't trigger a release.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -3,8 +3,12 @@ set -e
 
 echo "Running comprehensive checks before push..."
 
-# Full test suite with coverage
-npm run test:full
+# Full suite without coverage — mirrors the CI gate (the sharded `test` job runs
+# test:ci). Coverage is main-only and non-blocking in CI (the `coverage` job is
+# excluded from ci-pass), so gating pushes on it is stricter than CI and pays the
+# v8 instrumentation + report cost on every push. Set VITEST_MAX_WORKERS to trade
+# RAM for wall-time if you have the headroom (default 4; each fork caps at 6 GB).
+npm run test:ci
 
 # Detect unused code
 npm run knip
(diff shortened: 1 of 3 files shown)


andymai:
### Measured impact (this PR's own pre-push, dogfooded)

Same machine, back-to-back full-suite runs:

| Hook | Command | Wall time |
|------|---------|-----------|
| before | `test:full` (+ v8 coverage) | **180.3s** |
| after  | `test:ci` (no coverage)     | **71.4s** |

~2.5× faster (−60%) on ~3,300 tests. The ~109s delta is v8 coverage instrumentation + lcov/text report generation — work CI already runs main-only and excludes from `ci-pass`.

PR #1117, opened 2026-05-31
feat(kernel): Manifold mesh/CSG kernel adapter for fast previews

andymai:
## Summary

Adds a third registered geometry kernel, **`manifold`**, backed by Google's [Manifold](https://github.com/elalish/manifold) (`manifold-3d`, prebuilt WASM). It is a drop-in `KernelAdapter` positioned honestly as a **fast preview accelerator** — watertight mesh/CSG for interactive previews — while **exact B-rep export is preserved** by replaying a recorded operation graph onto the OCCT kernel.

The tagline the code embodies: **preview on mesh, export on B-rep.**

> Note: this deliberately goes against brepjs's B-rep-first positioning. The value is speed for previews without giving up B-rep export.

## Architecture (dual-kernel)

- Every manifold `KernelShape` is `{ manifold, node }`, where `node` is an `OpNode` recording the **exact construction intent** (op + params + inputs) plus a `replayable` flag (AND-propagated from inputs).
- **Native ops** (primitives, booleans, transforms, hull, extrude/revolve, tessellation, measurement, STL/OBJ/PLY/GLB IO) run directly on Manifold — instant.
- **B-rep-only ops** (fillet, chamfer, shell, loft, sweep, offset) render a fast **mesh approximation** for preview while recording exact intent.
- **B-rep export** (STEP/IGES/BREP/XCAF) and exact **NURBS curve/surface queries** replay the op-graph onto `getKernel('occt')` and answer from the real B-rep (memoized).
- **Degradation:** shapes tainted by a raw-mesh origin (`importGLB/OBJ/STL`, `meshBoolean`) are non-replayable → export degrades to faceted output with a `console.warn` instead of lying.
- **2D / constraint-sketch / projection** delegate to OCCT when registered, else throw a clear unsupported error.

## Packaging

- `packages/brepjs-manifold/` — thin loader exporting `initManifold()`; declared as an optional peer dep (mirrors `brepjs-opencascade`).
- Adapter lives in `src/kernel/manifold/` (mirrors the brepkit module layout).
- `initFromManifold(module)` registers it; exported from the public API.

## Testing

Tiered parity harness (`tests/parity/manifold.test.ts`, **104 tests**) comparing manifold ↔ OCCT:

- **Tier A (tight):** primitives/booleans/transforms/measurement — tight metric + Hausdorff bands.
- **Tier B (loose):** fillet/chamfer/shell — loose Hausdorff for mesh approximation vs B-rep.
- **Tier C (replay oracle):** for each replayable op, replay the manifold op-graph onto OCCT and assert it matches a **direct** OCCT build (B-rep vs B-rep, near-exact) — including subset edge-selection.
- **Tier D (degradation):** raw-mesh origin → faceted STEP + warning.

`TEST_KERNEL=manifold` selects it; informational coverage only (not in the strict cross-kernel agreement gate, as mesh-vs-B-rep divergence is expected by design).

## Unsupported by design (honest throws)

B-rep builders (`makeEdge`/`makeWire`/`makeFace`/`makeVertex`/arcs/axes), arena checkpoint/`executeBatch` (Manifold has no arena), 3MF IO (no zip dependency), per-edge/per-face modifier callbacks.

## Notable correctness work (post-review)

Two independent reviews surfaced and fixed: true volume-centroid `centerOfMass` (was AABB center); geometric **witness-point** edge/face selection so subset fillet/chamfer replay onto the correct OCCT sub-shapes (positional indices don't correspond across kernels); a `dispose()` double-free from pass-through ops aliasing one manifold solid; `draftPrism`/`revolve` contract fixes; STL vertex welding.

## Verification

`typecheck` / `lint` / `check:boundaries` / `knip` / `test:ci` (3329 passed) — all green. Manifold parity 104/104.

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -1,5 +1,5 @@
 [
-  { "name": "total (all JS)", "path": "dist/*.js", "limit": "265 kB" },
+  { "name": "total (all JS)", "path": "dist/*.js", "limit": "285 kB" },
   { "name": "brepjs", "path": "dist/brepjs.js", "limit": "55 kB" },
   { "name": "brepjs/core", "path": "dist/core.js", "limit": "2 kB" },
   { "name": "brepjs/result", "path": "dist/result.js", "limit": "1 kB" },
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,11 +11,13 @@
       "workspaces": [
         "packages/brepjs-opencascade",
         "packages/brepjs-bim",
+        "packages/brepjs-manifold",
         "apps/playground",
         "apps/docs"
       ],
       "dependencies": {
         "flatbush": "4.5.1",
+        "manifold-3d": "3.5.0",
         "opentype.js": "1.3.4"
       },
       "devDependencies": {
@@ -1883,7 +1885,6 @@
       "version": "1.10.0",
       "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.10.0.tgz",
       "integrity": "sha512-ewvYlk86xUoGI0zQRNq/mC+16R1QeDlKQy21Ki3oSYXNgLb45GV1P6A0M+/s6nyCuNDqe5VpaY84BzXGwVbwFA==",
-      "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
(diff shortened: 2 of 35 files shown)


andymai:
All four review findings are addressed in `cb8e5538` (the inline threads above re-anchored to the new commit but point at the now-fixed lines):

- **importGLB index width** — `ioOps.ts` now derives `stride` from the accessor `componentType` (UNSIGNED_INT/SHORT/BYTE) and sizes the `DataView` as `count * stride`, reading via `getUint32/16/8` accordingly. UINT16 GLBs no longer throw.
- **brepCache** — typed `WeakMap<OpNode, KernelShape>`; `dispose(handle)` now frees the replayed OCCT B-rep for that handle's node and removes the entry, so WASM shapes aren't stranded.
- **filletBatch per-edge radii** — replay maps `radii[i]` to `edges[i]` via a per-edge callback instead of applying `radii[0]` to all; Tier C oracle covers two edges with distinct radii.
- **STL/OBJ/PLY stride** — exporters read vertices at the mesh `numProp` stride via `vertexStride()`.

Greptile review check is green; all CI checks pass.

andymai:
The three inline threads above are stale re-anchors of round-1 findings — all are fixed in the current HEAD (`0bb7e721`):
- `importGLB` sizes the index DataView as `count * stride` from the accessor componentType (UINT16/UINT8 supported).
- `filletBatch` replay maps `radii[i]` to `edges[i]` via a per-edge callback.
- `vertexCount`/`vertexAt` use `vertexStride(mesh)` (numProp-aware).

The remaining 3/5 driver is intermediate OCCT WASM-shape accumulation in the replay engine on repeated B-rep export — preview path and correctness are unaffected. Tracked as #1118 for a proper disposal-scope pass with use-after-free verification (metric parity tests can't catch that, so it shouldn't be rushed here).

Merging: all required checks green.

Issue #1136, opened 2026-06-01
Migrate default init() kernel from brepjs-opencascade to occt-wasm


andymai:
## Motivation

The `brepjs-opencascade` OCCT build (used by the default `init()` → `'occt'` kernel) has a class of kernel-level crashes that the newer **`occt-wasm`** build does not. Concretely, the #1126 STEP-export OOB trap reproduces on `brepjs-opencascade` 0.16.0 but **`occt-wasm` 3.1.0 exports the identical shape to a valid 43,587-byte STEP** (verified with BREP-loaded inputs through the same `fuseAll` + export path). Upstream tracking: andymai/opencascade.js#3.

Migrating the default kernel to `occt-wasm` would sidestep #1126-class crashes without waiting on an OCCT/Docker rebuild of the legacy build, and consolidates on the build that's already "current".

## What this would involve

- Make the default `init()` register the `occt-wasm` adapter instead of importing `brepjs-opencascade`. Today `occt-wasm` requires explicit `registerKernel()` because its WASM loading uses Node.js APIs (`import.meta.resolve` / `locateFile`) — `src/kernel/index.ts` notes this. The loader needs a browser-safe path.
- Run the full conformance/parity suite with `TEST_KERNEL=occt-wasm` as the default and close divergences.

## Known blocker

`occt-wasm`'s frenet sweep currently errors: `sweepPipeShell:` (empty message) when building the #1126 handrail. Any migration must close this and other op gaps first — this is a sizable effort, not a drop-in swap. A gap audit of `occt-wasm` vs `brepjs-opencascade` should precede the switch.

## Acceptance

- Default `init()` uses `occt-wasm`; full suite green; #1126 (and similar) no longer crash.

andymai:
## occt-wasm STEP output verified geometrically faithful

Followed up on whether occt-wasm merely *avoids the crash* or actually produces correct geometry. It's faithful — verified two independent ways on the #1126 repro shape (`fuseAll([tread0, rail])`, 2 disjoint bodies):

1. **occt-wasm self round-trip** — export → re-import on occt-wasm.
2. **Cross-kernel** — occt-wasm exports the STEP, the `occt` (brepjs-opencascade) kernel independently re-imports it.

| Metric | Original | After STEP round-trip | Match |
|---|---|---|---|
| Solid count | 2 | 2 | ✅ |
| Tread volume | 3136.566105344 | 3136.566105345 | ✅ ~10 sig figs |
[...]

Issue #1126, opened 2026-05-31
exportSTEP rejects valid sweep+fuse compound ("Failed to read exported STEP file")


andymai:
## Summary

A geometrically-valid-looking shape built from `sweep` + `fuseAll` **verifies fine** (positive volume, sensible bounds, `validSolid` where checked) but `exportSTEP` returns `Result.Err("Failed to read exported STEP file")`. So a shape that all in-memory checks accept cannot round-trip to STEP — the primary interchange format.

## Repro

A miniature spiral-staircase model (one of the brepjs-agent benchmark parts, #1124):

- **Central column**: `cylinder`
- **Treads**: annular-sector profile drawn with `threePointsArcTo` (outer + inner arcs), `sketchOnPlane('XY').extrude(...)` → `ValidSolid`, arranged helically (~20 of them)
- **Balusters**: vertical `cylinder` rods at each tread's outer-mid radius
- **Handrail**: `sketchCircle(...).wire` swept along a helical path via `sweep(railProfile, railPath, { frenet: true })`
- Everything combined with `fuseAll([...])`

The resulting shape:
- ✅ `measureVolume` > 0, `getBounds` sane, renders correctly in the viewer
- ❌ `exportSTEP(shape)` → `Err("Failed to read exported STEP file")` (the writer's read-back validation rejects it)

## Suspected cause

The fused result is almost certainly a `Compound` containing a sub-shape that `STEPControl_Writer` (or the read-back step that validates the write) won't accept — likely a degenerate/near-degenerate face from the frenet sweep along the tight helix, or a non-manifold seam where the swept rail meets the balusters. The in-memory volume/bounds checks don't catch this because they don't exercise STEP's stricter topology/geometry validation.

## Ask

1. Investigate whether the offending sub-shape comes from the **sweep**, the **fuseAll**, or the **STEP writer** itself.
2. Consider running a validity/heal pass (or surfacing the specific invalid sub-shape) before/at export, and
3. At minimum, surface a **more actionable error** than "Failed to read exported STEP file" — e.g. which sub-shape index/type failed, so callers can locate and heal it.

I can attach the full `.brep.ts` repro on request (it lives in gitignored benchmark scratch). Surfaced by the brepjs-agent work (#1124).

andymai:
## Root cause (investigated)

Reproduced with the #09 spiral-staircase benchmark part and bisected the failure. Findings:

### 1. The error is a WASM trap, not a write or read failure
The raw error from the kernel is:
```
[...]

PR #1143, opened 2026-06-01
docs(decisions): add ADR-0013 voxel geometry domain

andymai:
## What

Adds **ADR-0013: Voxel Geometry Domain (PicoGK-inspired)** — the in-repo decision record for a native voxel / SDF geometry domain that strengthens B-rep weaknesses (non-watertight mesh repair, robust offset/shell, lattices/TPMS) as a first-class peer to B-rep.

Docs-only. No code yet — this records the architecture so implementation can start without open forks.

## Architecture: four stable seams

```
Grid → Ops → Contour → Bridge
```

v1 fills each seam minimally; every long-term capability is additive behind the *same* seam, so the minimal v1 grows to the full vision without restructuring:

| Seam | v1 | Long-term |
|---|---|---|
| Grid | dense array + active mask (behind an accessor abstraction) | hashed sparse blocks |
| Ops | boolean + non-watertight repair | offset, shell, lattice/TPMS, fields |
| Contour | surface nets | manifold dual contouring |
| Bridge | voxel→mesh (`KernelMeshResult`) | brep↔voxel, mesh→brep replay |

## Key decisions

- **Parallel domain, NOT a `KernelAdapter`** — voxel/SDF vocabulary doesn't fit the 14 B-rep sub-interfaces. Context via `getVoxel(id?)`, mirroring `getKernel(id?)`. (Note: `brepjs-manifold` *is* a `KernelAdapter`, so a true parallel domain is a new pattern.)
- **Native Rust → WASM**, shipped as a separately-published `brepjs-voxel-wasm` artifact (occt-wasm precedent) so root CI / the 4-shard matrix / pre-push stay Rust-free.
- **Independent sibling of brepjs-manifold** with a hard scope line; shares only the mesh type + an extracted replay helper.
- **v1 = the repair slice**, motivated by a non-watertight imported mesh (OCCT can't boolean non-watertight = the gating acceptance leg). Front-loads the hardest algorithm: mesh→SDF sign via Fast Winding Number.
- Layer assignment `src/voxel` (L2) + `src/lattice` (L3); `check-layer-boundaries.sh` must gain `voxel`/`lattice` arms (csg/gear/ns precedent).

## Note

Explicitly **not** a fix for #1126 — that's a build-specific BOPAlgo bug resolved by the #1136 migration; a voxel engine can't pass its exact-STEP acceptance by construction. #1126 is a counter-example, not motivation.

The second commit applies a research-driven correction: naive surface nets is watertight but not strictly 2-manifold/sharp, so acceptance is a runtime invariant test (strict 2-manifoldness waits for the manifold-DC contourer).

## Related

ADR-0001 (layering), ADR-0002 (kernel abstraction), ADR-0007 (interface segregation); `brepjs-manifold`; `brepjs-opencascade`/`occt-wasm` (published-wasm precedent); #1126, #1136.

### DIFF
--- a/docs/decisions/0013-voxel-domain.md
+++ b/docs/decisions/0013-voxel-domain.md
@@ -1,0 +1,211 @@
+# ADR-0013: Voxel Geometry Domain (PicoGK-inspired)
+
+**Status**: Proposed
+**Date**: 2026-06-01
+**Authors**: Andy Aragon
+
+## Context
+
+brepjs is an exact B-rep library. B-rep is precise and parametric but fragile on a class of
+problems that fail across _all_ exact kernels (OCCT, brepkit), not as a tuning issue:
+
+- booleans on non-watertight / self-intersecting **imported** meshes,
+- robust uniform offset / shell on concave geometry,
+- lattices / TPMS (gyroid, Schwarz, diamond) and graded infill,
+- organic / generative geometry, defeaturing, large-N CSG.
+
+A voxel / signed-distance-field (SDF) domain — the approach taken by PicoGK (C# → PicoGKRuntime
+C++ → OpenVDB) — solves these at the cost of exactness (resolution-bound, mesh output). This ADR
+records the architecture so implementation (starting with a P0 spike) can proceed without open
+forks. **No code exists yet.** The broader plan lives in the project owner's design vault; this ADR
+is the in-repo decision record.
+
+Note: this is explicitly **not** a fix for the #1126 STEP-export crash — that is a build-specific
+`BOPAlgo` bug resolved by the #1136 default-kernel migration. A voxel engine cannot pass #1126's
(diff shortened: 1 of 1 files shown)


andymai:
Thanks @greptile-apps — valid catch. Addressed in 59219ad.

§9 now enumerates the **complete** token set for all three locations (`get_layer()` arms, the printed legend, and the header comment) rather than just the additions, and explicitly flags the pre-existing drift to restore while adding `voxel`/`lattice`:

- **L2:** `topology, 2d, operations, query, measurement, io, worker, csg, voxel` — the legend currently omits `csg`; the header comment omits both `worker` and `csg`.
- **L3:** `sketching, text, projection, gear, ns, lattice` — the header comment also currently omits `ns` (one more than the review flagged).

This makes the P0 script edit self-contained so an implementer following the ADR literally produces an accurate legend/header.

PR #1146, opened 2026-06-01
feat(voxel): add voxel/SDF domain foundation

andymai:
## Summary

First increment of the **ADR-0013 voxel/SDF geometry domain** — a parallel sibling to the B-rep kernel. This is the **Foundation seam**: a Rust→WASM engine artifact, a TS engine registry, and the Generalized Winding Number (GWN) sign as the first real capability, wired end-to-end.

The GWN keystone was validated in a throwaway spike before this PR; here it's ported verbatim into the published-artifact crate and made callable across the wasm boundary.

## What's included

**`packages/brepjs-voxel-wasm`** — Apache-2.0 Rust crate (modeled on the `brepjs-opencascade` committed-artifact pattern)
- Validated FWN ported into `src/fwn.rs` (f64 internal for precision, f32 at the boundary to match `KernelMeshResult`)
- `wasm-bindgen` surface: `winding_numbers` / `points_inside` / `version`, flat typed-array calls (no zero-copy, per ADR)
- `crate-type = ["cdylib", "rlib"]` so the correctness assertions run on native `cargo test`
- **24 KB** committed wasm (budget was 60–200 KB); `npm run build` regenerates reproducibly

**`packages/brepjs-voxel`** — env-aware loader (`loadVoxelEngine`); Node reads wasm bytes via `createRequire` resolution, browser auto-fetches. Zero `brepjs` dependency (mirrors `brepjs-opencascade`).

**`src/voxel/` (Layer 2)** — the parallel-domain pattern ADR-0013 introduces
- `registry.ts` — `getVoxel(id?)` / `registerVoxel` / `initVoxel`, mirroring the kernel registry shape but independent of it
- `signFns.ts` — `windingNumbers` → `Float32Array`, `pointsInside` → `boolean[]`, with flat-buffer preconditions
- `engine.ts` — snake-case `VoxelEngine` interface the raw wasm module satisfies with zero glue
- surfaced through the public barrel

**Wiring** — both packages in `workspaces`; vitest alias for the artifact; `scripts/check-layer-boundaries.sh` gains `voxel`→L2 / `lattice`→L3 arms (also clearing the pre-existing legend/header drift ADR-0013 §9 flagged); knip ignore for the artifact.

## Verification

- `tests/voxel.test.ts` — end-to-end TS → wasm keystone: watertight cube exact, **holey cube center still classified inside** (the case ray-parity gets wrong), above-hole outside, bad-buffer + unknown-id throws. 5/5 under both occt + brepkit projects.
- `cargo test` 4/4 · typecheck · lint · boundaries · full `test:ci` (3350 tests) · knip — all green.
- Public-API surface guard updated with the 6 new exports.

## Next

The remaining v1 is the **Grid → Ops → Contour → Bridge** seams (dense grid + mask behind an accessor, edge normals, voxel boolean, hierarchical FWN, flood-fill, surface nets, `voxelsToMesh` bridge, GLB out).


andymai:
Addressed the round-2 Greptile feedback in `d365341`:

- **Out-of-bounds triangle index (P1):** `validateInputs` now bounds-checks every triangle index against `vertices.length / 3` before the call crosses into wasm. Previously an out-of-range index would panic in Rust and surface as a `RuntimeError` wasm trap, escaping the `Result<T, BrepError>` contract; it now returns `err(validationError('VOXEL_INVALID_TRIANGLE_INDEX', …))` like the other precondition failures. Added a test asserting `isErr` on `triangles=[0,1,100]` against an 8-vertex mesh.
- The restated "must return `Result`, not throw" comment is fully covered by this — with the trap path closed, `windingNumbers`/`pointsInside` have no remaining throw paths (validation + `getVoxel` errors are both returned as values).

6/6 voxel tests pass; typecheck, lint, boundaries, knip green.

PR #1149, opened 2026-06-01
feat(voxel): Grid/Ops/Contour/Bridge seams (v1 repair slice)

andymai:
## Summary

Completes the **ADR-0013 v1 repair slice** on top of the merged foundation (#1146): import a non-watertight triangle mesh → voxelize an FWN-signed SDF → Surface-Nets contour it back into a closed mesh. Each of the four seams is filled minimally, with optimizations deferred *behind* the same abstractions.

## Seams

**Rust (`packages/brepjs-voxel-wasm`)**
- **Grid** (`grid.rs`) — dense SDF grid behind an accessor abstraction (so a sparse grid can swap in later, no raw indexing exposed); `MAX_VOXELS` cap → `GridError` via `checked_mul` (no OOM).
- **Ops** (`ops.rs`) — `voxelize_mesh`: exact closest-point-on-triangle distance (Ericson Voronoi regions, f64) × FWN sign (inside → negative), so **holey input yields a watertight SDF**; voxel boolean `union`/`intersection`/`difference` (min/max/max-neg).
- **Contour** (`contour.rs`) — Surface Nets via `fast-surface-nets` 0.2 behind a `Contourer` trait (manifold dual contouring slots in behind it later); maps vertices into world coordinates.
- **Bridge** (`lib.rs`) — wasm `repair_mesh` → `RepairResult`. 36 KB artifact (cap 250 KB).

**TS**
- `src/voxel/repairFns.ts` — `repairMesh(mesh, opts?) → Result<KernelMeshResult>`. Validates the mesh (incl. in-bounds indices), `resolution`, and `padding`, and **wraps the fallible wasm call in try/catch** so a grid-cap error returns `err(...)` rather than escaping as a thrown wasm exception. Surfaced through the barrel; public-API allowlist updated.

## Verification
- **20 cargo tests** (grid/ops/contour + the FWN keystone), **73 vitest** (incl. end-to-end holey-cube repair + Result-contract guards: OOB index, empty mesh, bad resolution, and the voxel-cap path returning `err` not a trap).
- typecheck, lint, boundaries, knip — all green.

## Build process
Built via a multi-agent workflow (parallel survey → sequential per-seam implementation, each self-gated with `cargo`/`wasm-pack`/`vitest` → integrate → parallel review). The review surfaced a `Result`-contract breach (the fallible `repair_mesh` call lacked a try/catch); fixed here with the validation + try/catch above and 4 added guard tests.

## Deferred behind the seams (out of v1 scope)
Narrow-band + BVH/Barnes-Hut hierarchical FWN, flood-fill, sparse grid, manifold dual contouring, and a raw-`KernelMeshResult`→GLB adapter (the existing `exportGlb` takes `ShapeMesh`, not a bare mesh — no path exists today).


andymai:
Addressed all three review comments in `4941f74`:

- **WASM RepairResult lifecycle (P1):** `VoxelRepairResult` now forwards `free()` and `[Symbol.dispose]()`, and `repairMesh` binds the handle with `using` so the WASM allocation is released on scope exit (including on a thrown mid-extraction). The getters copy their buffers out, so the returned `KernelMeshResult` stays valid after `free()` — the end-to-end repair test confirms this.
- **Padding clipping (P1):** `padding` must now be `>= 1` (Surface Nets needs a ≥1-voxel air margin); `{ padding: 0 }` returns `err`, asserted in the test.

73 vitest + 20 cargo green; typecheck/lint/boundaries/knip clean.

PR #1152, opened 2026-06-02
feat(lattice): add Layer-3 TPMS lattice domain

andymai:
## Summary

Adds the **`src/lattice` (Layer 3) TPMS lattice domain** on top of the voxel seams — the PicoGK-style "lightweight a part with a printable lattice" capability. Triply-periodic minimal-surface lattices (gyroid, Schwarz-P, diamond) and lattice infill of a mesh region.

This batch is almost pure *reuse* of the Grid/Ops/Contour/Bridge seams — the only genuinely new Rust is the TPMS field generator; infill is `voxel_intersection(solidSDF, tpmsShell)` followed by the existing surface-nets contour.

## API

```ts
latticeInfill(mesh, { type: 'gyroid' | 'schwarzP' | 'diamond', period, thickness }) → Result<KernelMeshResult>
tpmsLattice({ min, max }, opts) → Result<KernelMeshResult>
```

## Implementation

**Rust (`packages/brepjs-voxel-wasm`)**
- `tpms.rs` — `LatticeType` + `tpms_value` (standard nodal trig, coords scaled by 2π/period) + `fill_tpms_shell` (`|f| - thickness/2`, negative = strut material, matching the inside-negative SDF convention)
- `lib.rs` — `lattice_infill` (voxelize mesh → FWN-signed solid SDF → ∩ TPMS shell → surface-nets) and `tpms_box` (box-clipped lattice); both reuse the `RepairResult` handle
- `ops.rs` — `fill_box_sdf` (exact AABB box SDF) so `tpms_box` clips the periodic lattice cleanly at its bounds

**TS (`src/lattice`, Layer 3)** — `latticeInfill` / `tpmsLattice` → `Result<KernelMeshResult>`, mirroring `repairFns`: validate-before-wasm (mesh, period>0, thickness>0, resolution≥1, padding≥1, bounds), `try/catch` the fallible wasm call, `using` the memory-owning handle.

## Verification
- **26 cargo + 84 vitest** (infill all 3 lattice families, box-clip bounds check, Result-contract guards: bad period/thickness/padding/indices/bounds → `err`)
- typecheck, lint, boundaries, knip green; size 285.94 kB < 290 kB; 48 KB wasm

## Build process
Built via a multi-agent workflow (survey → sequential Rust→TS implement → integrate → review). The review found one **P2** — `tpms_box` wrote shell material into the padding ring, so the box-lattice ran past `[min..max]` (`latticeInfill` was unaffected — the solid intersection clamps it). Fixed with `fill_box_sdf` + a bounds-check test that closes the test-gap which let it hide.

## Deferred (behind the seams)
`latticeInfillShape(solid, …)` convenience (mesh a B-rep solid then infill), beam/strut + graded lattices, a playground demo.


andymai:
Addressed both review comments in `0c53603`:

- **`period > 0` guard at the wasm boundary (P2):** added `check_lattice_params` (rejects non-finite/non-positive `period` and `thickness` with a `JsError`) to both `lattice_infill` and `tpms_box`. The TS layer already validates these, but the wasm artifact is consumed directly too, and `period == 0` makes the `2π/period` scale NaN — so this is defense-in-depth at the boundary.
- **Periodicity coverage (P2):** the Rust test now verifies all three lattice families (gyroid/Schwarz-P/diamond) repeat after one period on every axis (x/y/z), not just gyroid-on-x.

26 cargo + 84 vitest green; typecheck/lint/boundaries/knip clean.

PR #1154, opened 2026-06-02
feat(voxel): brep↔voxel interop — robust CSG, offset/shell, shape conveniences

andymai:
## Summary

Adds **voxel-based mesh operations that compose with B-rep shapes**, on top of the voxel seams. This directly attacks two of brepjs's stated weaknesses — **robust CSG** and **offset/shell** — by routing through a representation (a dense SDF) where those operations are *unconditionally* well-defined: `min`/`max` of two fields can't fail, and an iso-level shift can't self-intersect.

## API

**Layer 2 (`src/voxel`)** — mesh ops, all `Result<KernelMeshResult>`:
```ts
offsetMesh(mesh, distance, opts?)              // exact SDF iso-shift (+grow / -shrink)
shellMesh(mesh, thickness, opts?)              // hollow inward wall
voxelBoolean(a, b, 'union'|'intersection'|'difference', opts?)   // robust CSG
```
**Shape conveniences** (mesh a B-rep shape, then the op):
```ts
offsetShape(shape, d) · shellShape(shape, t) · voxelBooleanShapes(a, b, op)
latticeInfillShape(shape, { type:'gyroid', period, thickness })   // the "lattice this part" one-liner (Layer 3)
```

## Implementation

**Rust (`packages/brepjs-voxel-wasm`)** — three wasm entries reusing `voxelize_mesh` + `ops` + `surface_nets_mesh`:
- `offset_mesh` — `new_sdf = sdf - distance` (the voxelizer is a true SDF, so this is exact); the bbox is expanded by `distance.max(0)` so an outward offset doesn't clip at the grid edge
- `shell_mesh` — `max(solid, -(solid + thickness))`
- `voxel_boolean` — voxelize both meshes onto a shared union-bbox grid, then `min`/`max`/`max(a,-b)`
- `ops.rs` — `offset_sdf` + `shell_sdf` helpers

**TS** — `shapeToMeshInput` meshes a Shape via the topology `mesh()` API (`ShapeMesh.vertices/.triangles` map field-for-field to `VoxelMeshInput`); every op mirrors `repairFns` (validate-in-TS, defense-in-depth wasm guards, `try/catch` the fallible call, `using` the handle).

## Verification
- **31 cargo + 97 vitest** — offset (both directions, exact bbox), shell, all 3 booleans, the B-rep conveniences against a real OCCT box, and Result-contract guards (non-finite distance, non-positive thickness, out-of-bounds indices)
- typecheck, lint, boundaries, knip green; size 287.15 kB < 290 kB; 52 KB wasm

## Build process
Built via a multi-agent workflow; the review returned **no P1/P2 findings** (two P3 comment nits and a test-symmetry gap addressed here).

## Deferred
Performance scaling (hierarchical FWN — the brute-force voxelizer is O(voxels×triangles)), persistent grid handles, meshToBrep reconstruction.


andymai:
Addressed both review comments:

- **Empty-result silently returns `ok` (P2):** `meshFromResult` now returns `Result` and errs with `VOXEL_DEGENERATE_RESULT` when an op produces no geometry (an over-shrunk offset, a disjoint intersection), mirroring the input-side `VOXEL_EMPTY_MESH` guard. Added a test: intersecting two disjoint cubes returns `err`.
- **Index guard (P2):** rebuilt the shifted-cube fixture with `Float32Array.from(VERTS, …)` — no index access at all. (Note: under this repo's `noUncheckedIndexedAccess`, `Float32Array` indexing widens to `number | undefined`, so the original guard wasn't strictly dead; the functional form sidesteps it cleanly without a cast.)

98 vitest + 31 cargo green.

PR #1163, opened 2026-06-02
fix(occtwasm): expand wireframe polylines into a line list in meshEdges

andymai:
## Problem

occt-wasm's `meshEdges` produces malformed edge geometry. For an identical bin, it emits ~3× the edge segments opencascade does (3568 vs 1256) including ~1168 zero-length pairs. Downstream this renders as **dashed curved edges and spurious line segments cutting across empty space** — visible in the normal shaded view of any bin, especially around wall cutouts (many short edges).

## Root cause

`wireframe()` samples each edge into a **polyline** (N points along the curve), concatenated per edge with `edgeGroups` giving each edge's float range. But `KernelEdgeMeshResult.lines` is contracted to a flat **line list** — consecutive point *pairs*, one per segment, for `THREE.LineSegments` — which is what the `manifold` and opencascade kernels produce. `meshEdges` returned the raw polyline points, so a consumer pairing them as a line list:

- draws only every other segment of each curve (dashed edges),
- joins the **end of one edge to the start of the next** → a spurious segment (zero-length where they share a vertex, a diagonal across space where they don't).

## Fix

Expand each per-edge polyline into consecutive point pairs (`pushPolylineSegments`), and drop degenerate samples the curve sampler repeats on tiny edges. Output now matches the line-list contract; verified against opencascade's clean output. Full occt suite green (2979 passed).

### DIFF
--- a/src/kernel/occtWasm/meshOps.ts
+++ b/src/kernel/occtWasm/meshOps.ts
@@ -128,7 +128,7 @@ export function meshEdges(
     const lineList: number[] = [];
     const edgeGroups: Array<{ start: number; count: number; edgeHash: number }> = [];
     const egCount = edgeData.edgeGroupCount;
-    const egPtr = edgeData.getEdgeGroupsPtr() >> 2;
+    const egPtr = egCount > 0 ? edgeData.getEdgeGroupsPtr() >> 2 : 0;
     for (let i = 0; i < egCount; i += 3) {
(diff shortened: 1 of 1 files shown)


andymai:
Addressed — `getEdgeGroupsPtr()` is now guarded behind `egCount > 0` (returns 0 ptr otherwise; the loop is a no-op), matching `mesh()`'s convention.

PR #1172, opened 2026-06-02
feat(sheetmetal): add experimental sheet-metal domain package

andymai:
## What

Adds `packages/brepjs-sheetmetal` — a programmatic **sheet-metal CAD domain** for brepjs, as an experimental, unpublished satellite package (same posture as `brepjs-bim`/`brepjs-manifold`).

Motivated by a user asking whether brepjs could do *"miter cuts on bent profiles before unfolding"* — the exact author→miter→unfold workflow this implements.

## Phase-1 vertical slice (straight/cylindrical bends, OCCT-WASM)

- **Authoring** — base flat + flanges folded up about a recorded bend axis, joined by real cylindrical bend patches. Public-API-only construction.
- **Unfold** — feature-tree-driven analytic flattening to a true 2D flat pattern. Each flange develops **perpendicular to its own bend line** (cross/L development), with bend allowance `BA = (π/180)·angle·(R + K·T)`.
- **Auto corner-miter** — recorded into the model and replayed as a 2D corner notch, so a mitered corner actually appears in the unfolded outline (a diagonal edge), not just on the 3D solid.
- **Outputs** — annotated multi-layer DXF (package-local writer; the core writer is R12 LINE/POLYLINE-only), 2D blueprint outline, folded 3D solid (STEP/GLB via core IO), and a JSON bend report.
- **Materials & validation** — starter steel/aluminum gauge library; manufacturability warnings (solid validity, flange collision, min bend radius) returned alongside the result.
- **API** — three-layer (`*Fns` → `api` → fluent facade), matching brepjs convention.

## Design note: the kernel pivot

An adversarial design review caught that the obvious unfold path — reading cylinder geometry via `getSurfaceCylinderData` — is unexported, returns no axis, **and throws `notImplemented()` on OCCT-WASM**. Because Phase 1 *authors* parts (rather than unfolding foreign solids), the bend axis/radius/angle are known inputs recorded in a feature tree, so the unfold reads the tree, not the B-rep surface. This keeps the package on the public, OCCT-WASM-safe API with no kernel changes (how Fusion-style unfold works: consume feature history).

## Verification

- 66 tests vs OCCT-WASM (typecheck/lint/test green); dedicated `packages-sheetmetal` CI job wired into `ci-pass`.
- Unfold correctness is anchored by analytic bend-allowance assertions across K∈{0.33,0.44,0.5}, an independently-derived area-conservation oracle, and a reference bend-allowance table.
- Acceptance tests for the cross-development and miter-in-outline are **sabotage-proven** (they fail under a colinear layout).
- Standalone snapshot harness (`npm run snapshot`, no playground/GPU dep) renders folded + flat side-by-side and exports STEP; developed area matches the hand calculation exactly.

## Known Phase-1 limitations (documented, deferred)

- Closed/box parts: seam-cut path exists but authored parts form a star graph, so it's currently untested.
- Down-bends, conical/compound bends, and richer DXF polylines are out of scope for Phase 1.
- Flat-pattern→fold (the inverse authoring direction) is deferred to Phase 2.


andymai:
Thanks for the thorough multi-round review — addressed the concrete findings across the last several commits (DXF table count + R2000 conformance, duplicate-side guard, `SEAM_CUT` warning code, `CornerMiter` export, builder memoization, `flatLength` semantics, and the zero-gap miter crash), each with a regression test.

On the remaining **WASM-disposal** item: intentionally not changing it, for consistency with the rest of the codebase rather than as an oversight —

- This package uses only the **functional branded-type API** (`box`/`cylinder`/`fuse`/`cut`/`rotate`/`translate`), never `createHandle`/`createKernelHandle`. Core `src/operations` and `src/topology` chain these same functional booleans without manually disposing intermediates.
- brepjs reclaims undisposed handles via a **`FinalizationRegistry` safety net** (`src/core/disposal.ts`), so these are GC-reclaimed, not session-lifetime leaks.
- The repo's pattern checker (`scripts/check-patterns`, rule `require-using-for-handles`) scopes the `using` requirement to `createHandle`/`createKernelHandle` calls only — it does not flag functional-API intermediates — and the package passes it.

Adding explicit `DisposalScope`/`using` only here would make this satellite the lone divergence from every other domain. If deterministic disposal of functional intermediates becomes a codebase-wide convention, this package should adopt it alongside `operations`/`topology` rather than ahead of them. Merging as an experimental Phase-1 satellite per the "safe to merge" verdict.

PR #1177, opened 2026-06-03
feat(sheetmetal): multi-bend authoring, down-bends, closed-box seams

andymai:
## Phase 2 · PR1 of a stacked sequence

First PR of the maximal sheet-metal expansion. Generalizes the domain past the Phase-1 two-flange envelope and lays the load-bearing foundation (recursive tree-walk unfold) that the rest of the stack builds on.

### What's new
- **Authoring**: flanges off all four base edges; **chained flange-off-flange** bends (U-channels, Z-profiles, box walls); **up/down** fold direction; **partial/offset** flanges (multiple flanges per edge), with overlap rejection.
- **Unfold**: replaces the hardcoded `east`/`north` two-slot `RunLayout` with a **recursive BFS tree-walk** 2D placement, emitting a rectilinear-union outline for arbitrary arm trees (combs, trays, chains) plus bend lines and developed area.
- **Closed-box seams end-to-end**: authored `seam::` edges are excluded from the spanning tree and recorded as `SEAM_CUT`, so the cycle-closing edge stays unfolded and the part flattens into a valid connected pattern.
- **Propagation**: report (N bends), validate (real folded AABBs for chained flanges), miter, facade (`.seam()`), index exports (`FlatSide`, `SeamSpec`).
- **Harness**: parameterized into bracket + U-channel + tray demos.

### Verification
- typecheck / lint / build clean; **84 tests** (74 prior + 10 new) green against OCCT-WASM.
- New tests cover: chained U-channel + area invariant, opposite-edge U, 4-side tray, down-bend (below-plane bounds + volume), partial/offset flanges + overlap rejection, closed-box `SEAM_CUT`, multi-bend report.
- Built via a multi-agent Workflow (implement → adversarial review → fix). Review caught a masked **seam-BFS** correctness bug (a seam edge could be promoted into the spanning tree, demoting a real bend to the cut); fixed by partitioning seam edges before BFS.

### Scope notes
- The unfold outline is exact for the supported families (single flange/side, U-channels, trays, chains, combs); the 45° corner-miter chamfer applies to the classic base-L pair, other recorded miters cut the 3D solid but fall back to the un-chamfered union outline.
- `validateFns` collision is analytic/advisory (kernel-free AABB).

### Stack
Next PRs (sequential, merged into `main`): flat→fold inverse · corner/bend reliefs · cutouts/holes · tabs/slots/form features · contour/lofted flanges · foreign-solid unfold.

OCCT-WASM-first; brepkit parity deferred to a later sweep.

### DIFF
--- a/packages/brepjs-sheetmetal/README.md
+++ b/packages/brepjs-sheetmetal/README.md
@@ -10,21 +10,27 @@ Pipeline: **author part → auto-miter corners → fold to 3D → unfold to flat
 
 ## Scope
 
-First build is a single end-to-end vertical slice for straight (cylindrical) bends, authoring our own
-parts (not unfolding foreign solids). The bend model is K-factor based with an extensible schema, so it
-can grow toward bend-tables without a rewrite.
+Straight (cylindrical) bends, authoring our own parts (not unfolding foreign solids). The bend model is
+K-factor based with an extensible schema, so it can grow toward bend-tables without a rewrite.
 
 - Bend allowance: `BA = (π/180)·|angle|·(R + K·T)`
 - Defaults: units are mm; K-factor `0.44`; inner radius `= thickness`.
 
+Authoring supports arbitrary bend trees: flanges off any of the four base edges, **chained
+flange-off-flange** bends (U-channels, Z-profiles, box walls), **up/down** fold direction, and
+**partial/offset** flanges (more than one flange per edge). Closed profiles (tubes/boxes) author a seam
+edge that the unfold leaves uncut as a `SEAM_CUT`, flattening into a valid connected pattern.
+
 ## Status
 
-| Phase        | Contents                                                                                                             |
-| ------------ | -------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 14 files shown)


andymai:
Thanks — addressed all three findings in 24c1b5af:

1. **Seam-id fragility** — flange ids containing `::` are now rejected at authoring (it's the reserved `seam::<parent>::<child>` delimiter), and the parse uses `lastIndexOf` as defense-in-depth. So the parent/child pair can no longer be ambiguous.
2. **Unvalidated seam params** — `SeamSpec.angleDeg` (must be in (0, 180]) and `rule.innerRadius` (>= 0) are now validated before the seam bend reaches `part.bends`, matching the flange guards.
3. **Unreachable default** — replaced with `side satisfies never` so a future `FlatSide` variant is a compile error rather than a silent wrong `edgeOrigin`.

Added 3 regression tests (now 87 total).

andymai:
Fixed in 360a0429 (P1): `authorPart` now rejects a flange reusing the reserved base-flat ids `'root'` (= `ROOT_FLAT_ID`) or `'face-0'` at the id guard — so it can no longer overwrite the base frame and silently return wrong geometry. Added parameterized regression tests (now 89 total).

PR #1180, opened 2026-06-03
feat(sheetmetal): flat-pattern→fold inverse with non-circular round-trip oracle

andymai:
## Phase 2 · PR2 of a stacked sequence

Adds `fold` — the inverse of `unfold` — turning a flat pattern back into a 3D sheet-metal part, and establishes a **genuine round-trip oracle** that becomes the regression net for the rest of the stack.

### What's new
- **`fold(input: FlatInput)`** folds a `FoldRegion` tree up into a `SheetMetalPart`, reusing `authorPart`'s forward bend geometry (no duplicated bend math).
- **`patternToFlatInput(pattern, { thickness, ruleFor })`** recovers the region tree from the **2D flat-pattern geometry alone** — reading coordinates back out of the outline `Wire` and bend-line `Edge`s via `getEdges` / `curveStartPoint` / `curveEndPoint`. Regions, sides, offsets, spans, lengths, angles and directions are all derived from geometry, recovered in each parent's *local* frame (so a chained `return` folds off the correct edge). Only the bend **rule** (K-factor) is a supplied input — a flat pattern physically can't encode it.
- `partToFlatInput` routes through `unfold → patternToFlatInput`. Surfaced via `api.ts` / `facade.ts` (`foldFlat`) / `index.ts`.

### Why the oracle matters (and is non-circular)
The first cut of this PR was a circular no-op: `fold` called `authorPart` and the "oracle" read the region values back out of the feature tree, so the round-trip passed *by construction*. That was caught in adversarial review and reworked. Now `author → unfold → patternToFlatInput → fold` forces the data through **real 2D geometry**, so a bug in either direction breaks the volume assertion:
- Round-trip Δvol ~1e-13 mm³ across single bracket, U-channel (chain), 4-sided tray, down-bend, and partial/offset families; valid solids; matching bend/flange counts.
- **Perturbation tests** confirm the assertion has teeth (halving a recovered length/width diverges volume by >1 mm³).
- A dedicated test asserts the recovered `side`/`offset`/`span`/`length` equal the authored values (not just that volume matches).

### Quality
- `fold` validation reuses the canonical `validatePart` (COLLISION/INVALID_SOLID/MIN_RADIUS) instead of a hand-rolled copy; `edgeBasis2` is exported from `unfoldFns` and shared so `fold`/`unfold` stay in lockstep.
- typecheck / lint / build clean; **103 tests** (89 prior + 14 new) green against occt-wasm; snapshot harness reports the fold round-trip Δvol per demo.
- Built via a multi-agent Workflow (implement → adversarial review → fix); the circular-oracle defect was found and corrected before this PR was opened.

### Stack
PR1 (multi-bend authoring) merged as #1177. Remaining: corner/bend reliefs · cutouts/holes · tabs/slots/form features · contour/lofted flanges · foreign-solid unfold. OCCT-WASM-first.

### DIFF
--- a/packages/brepjs-sheetmetal/README.md
+++ b/packages/brepjs-sheetmetal/README.md
@@ -27,10 +27,20 @@ edge that the unfold leaves uncut as a `SEAM_CUT`, flattening into a valid conne
 | --------------- | -------------------------------------------------------------------------------------- |
 | Authoring       | 4-edge flanges, chained bends, up/down, partial/offset flanges, closed-box seams       |
 | Unfold          | recursive BFS tree-walk → rectilinear-union flat pattern + bend lines + developed area |
+| Fold            | `FlatInput` region-tree → 3D part (inverse of unfold); round-trips `unfold(fold)`      |
 | Miter / outputs | auto corner-miter, multi-layer DXF, JSON bend report, manufacturability warnings       |
 | API             | functional `*Fns` → short-named `api.ts` → fluent `sheetMetal()` facade                |
 
-`FlatInput` (flat-pattern → fold, the inverse direction) is not yet implemented.
+`fold(input: FlatInput)` folds a flat pattern back up into a 3D part — the inverse of `unfold`. A
+`FlatInput` is a region-tree (a base rectangle plus child fold regions, each with a fold line, angle,
+direction, and bend rule), the inverse of the unfold layout. `patternToFlatInput(pattern, { thickness,
+ruleFor })` recovers that region-tree from the **2D flat-pattern geometry alone** — the developed outline
+wire and bend-line edges — reading real coordinates back out with the public `getEdges` / `curveStartPoint`
+/ `curveEndPoint` readers; only the bend rule (which a flat pattern cannot encode) is a supplied input.
+`partToFlatInput(part)` chains `unfold` → `patternToFlatInput`, so `fold(partToFlatInput(part))` round-trips
+volume, validity, and bend/flange counts through the 2D geometry — making the round-trip a genuine,
+non-circular oracle. Fold reuses the forward authoring bend geometry wholesale (no duplicated bend math),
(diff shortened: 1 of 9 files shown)


andymai:
Addressed the review:

1. **`orderedLoop` overwrite** — now bails (`undefined` → `OUTLINE_NOT_CLOSED`) when two segments share a start-vertex key, instead of silently overwriting and tracing a wrong polygon.
2. **Hardcoded `1e-3` base probe** — replaced with a probe halfway to the first *positive* x grid line: scale-invariant (no unit assumption) and correct even with `xmin`/`ymin` flanges adding negative coords (the literal `(xs[0]+xs[1])/2` would land in the xmin strip for those parts). Added a sub-mm regression test (now 104 tests).
3. **`getEdges` disposal** — intentionally left as-is: this matches the established repo-wide convention for functional shape readers (core `src/sketching|operations|topology|query` and the package's own merged `dxfFns.ts` all use `getEdges` without manual disposal; the `disposal-required`/`require-using-for-handles` rule covers only `createHandle`/`createKernelHandle`). Disposing here would make the package inconsistent with core.

andymai:
The latest Greptile pass re-surfaced the prior three findings, but two of them are already resolved in 1e436a9 (it appears to have re-emitted against the cumulative diff):
- **orderedLoop overwrite** — fixed: `from.has(k)` guard now returns `undefined` before any overwrite (foldFns.ts:324).
- **baseRect `1e-3` probe** — fixed: now probes `xFirstPositive/2` (foldFns.ts:388), scale-invariant, no unit assumption.
- **getEdges disposal** — by design, consistent with core `src/` and the package's `dxfFns.ts`; the `disposal-required` rule covers only `createHandle`/`createKernelHandle`.

CI is green and there are no new concrete defects, so merging at the no-concrete-defect bar.

PR #1182, opened 2026-06-03
feat(sheetmetal): corner & bend reliefs for manufacturable multi-bend parts

andymai:
## Phase 2 · PR3 of a stacked sequence

Adds **relief notches** so multi-bend parts can actually be fabricated — without the material tearing or colliding at corners and partial-bend ends. This is the first PR that encodes *fabrication* knowledge on top of the geometry.

### What's new
- **Bend reliefs**: `addBendRelief(part, flangeId, spec)` + `autoBendReliefs(part)` cut a relief slot at each mid-edge end of a partial/offset bend (the ends where the bend line meets un-bent metal and would otherwise crack).
- **Corner reliefs**: `cornerRelief(part, idA, idB, spec)` notches the shared corner of two flanges so they clear when folded — the notch alternative to a 45° miter. It records a zero-gap miter so the `validate` `COLLISION` warning clears.
- Recorded features (`SheetMetalPart.reliefs`) cut into the 3D solid **and** replayed as 2D notches in `unfold`, mirroring the existing miter machinery. `developedArea` drops by the removed notch area.

### Design
- `ReliefSpec { shape: 'rectangular' | 'obround'; width?; depth? }` (obround records intent; geometry is rectangular in this version — documented). Defaults: `width ≈ thickness`, `depth ≈ developedLength + thickness`.
- `FlatPattern.bendLines` now carry the bend `id`, so reliefs locate the correct line even when multiple flanges share span/angle/direction (fixed a notch-misplacement bug found in review).
- Inward notch direction is derived from the placed child frame (correct for chained flanges, not just root). Both relief ops guard against severing the solid (`RELIEF_SEVERED_SOLID`).
- `fold` re-applies `FoldRegion.bendRelief`; the strict geometric round-trip oracle skips relief'd parts (a notched outline isn't a plain rectangle — same handling as mitered parts), with a feature-level volume check instead.

### Verification
- typecheck / lint / build clean; **115 tests** (104 prior + 11 new) green against occt-wasm; snapshot adds bend-relief + corner-relief demos.
- Geometric assertions: 3D volume drop ≈ notch volume, 2D `developedArea` drop ≈ notch area, notch placement on the correct bend line inside the parent, `COLLISION` resolved after corner relief, valid solids + valid closed outline wires.
- Built via a multi-agent Workflow (implement → adversarial review → fix); review caught the same-span notch-misplacement bug and a missing severed-solid guard, both fixed before opening.

### Stack
PR1 (multi-bend) #1177, PR2 (flat→fold + round-trip oracle) #1180 merged. Remaining: cutouts/holes · tabs/slots/form features · contour/lofted flanges · foreign-solid unfold. OCCT-WASM-first.

### DIFF
--- a/packages/brepjs-sheetmetal/.claude/agent-memory-local/engineer/MEMORY.md
+++ b/packages/brepjs-sheetmetal/.claude/agent-memory-local/engineer/MEMORY.md
@@ -1,0 +1,5 @@
+# Engineer Memory — brepjs-sheetmetal
+
+## Reliefs
+
+- [PR3 reliefs](project_sheetmetal_reliefs_pr3.md) — bend/corner relief invariants: bendLines carry id+inward, severed-solid guard, obround = recorded intent only, corner side = width??depth
--- a/packages/brepjs-sheetmetal/.claude/agent-memory-local/engineer/project_sheetmetal_reliefs_pr3.md
+++ b/packages/brepjs-sheetmetal/.claude/agent-memory-local/engineer/project_sheetmetal_reliefs_pr3.md
@@ -1,0 +1,20 @@
+---
+name: sheetmetal-reliefs-pr3
+description: brepjs-sheetmetal PR3 (feat/sheetmetal-reliefs) review-fix conventions and relief geometry gotchas
+metadata:
+  type: project
+---
+
+PR3 = bend + corner reliefs for brepjs-sheetmetal (branch feat/sheetmetal-reliefs).
+
+**Why:** Multi-bend parts tear/collide; reliefs are recorded features (like CornerMiter) cut from the 3D solid and replayed by unfold as 2D notches.
+
+**How to apply (relief geometry invariants):**
+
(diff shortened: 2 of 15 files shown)


andymai:
Addressed the review:

**P2s fixed (commit pending):**
- **autoBendReliefs double-work** — now delegates to `addBendRelief` directly and treats `BEND_RELIEF_NOT_NEEDED` as `continue`, removing the redundant geometry recompute and the throwaway tool solids on the check path.
- **Corner-relief misleading `depth`** — now records `depth = side` (the actual square side) so `width`/`depth` reconstruct the real cut.

**P1 disposal findings — intentionally not changed.** These match the established, deliberate convention across both core and the already-merged package code: the merged `miterFns.ts` builds the identical `box → translate → rotate` half-space tool chains with no manual disposal, `authorFns.ts` has none, and core `operations/booleanFns.ts` doesn't dispose functional intermediates either. brepjs relies on the `FinalizationRegistry` safety net in `core/disposal.ts` for handles not created via `createHandle`/`createKernelHandle` (the only handles the `require-using-for-handles` pattern rule covers). Disposing only in `reliefFns` would make it inconsistent with the rest of the package and core. (The autoBendReliefs simplification above does reduce the throwaway allocations on the hot path.) This is the same codebase-wide convention question resolved in the Phase-1 PR (#1172).

PR #1184, opened 2026-06-03
feat(sheetmetal): cutouts & holes that survive fold/unfold

andymai:
## Phase 2 · PR4 of a stacked sequence

Adds 2D **cutout features** — holes, slots, and polygons punched in a flat region — that appear in the 3D solid, the developed flat pattern, and the DXF, and survive the fold/unfold round-trip.

### What's new
- `addCutout` / `addHole` / `addSlot` / `addPolygonCutout` (`cutoutFns.ts`) place a feature in a named region's **local 2D coordinates** (`region` = a flange id, or `base`/`root`).
- `CutoutSpec` discriminated union (`hole | slot | polygon`); `CutoutFeature`; `FlatPattern.holes: Wire[]`; `SheetMetalPart.cutouts?`; `FoldRegion.cutouts?` + `FlatInput.baseCutouts?`.
- `unfold` emits one closed interior `Wire` per cutout and drops `developedArea` by the cutout areas (outer outline unchanged); `dxfFns` writes them on a new `CUTOUT` layer; `fold` re-applies them.

### Correctness — 2D↔3D coordinate consistency
The headline requirement: a hole authored at flange-local `(x, y)` must be cut on the **folded flange face** in 3D *and* reappear at the **matching spot** in the developed flat pattern.
- The 3D cut uses the region's world `FlatFrame` (`worldFrames` reconstructs it via the same `computeChildFrame` math `authorPart` uses) — verified: a hole on a 90° flange produces the same volume drop as on the flat base.
- The 2D loop uses the region's developed `Frame2` with the *same* local convention — verified: developed-loop centroid equals `developedOrigin + x·u + y·v`.
- Round-trip Δvol ~1e-12 mm³ for base and flange holes; both reviewers independently confirmed the frame consistency.

### Guards & quality
- `CUTOUT_OUT_OF_BOUNDS`, `CUTOUT_SEVERED_SOLID`, `UNKNOWN_REGION`; `attachCutouts` fails loud (`CUTOUT_REGION_UNMAPPED`) rather than silently dropping an unmappable cutout (would corrupt the refold volume).
- typecheck / lint / build clean; **130 tests** (115 prior + 15 new) green against occt-wasm; snapshot `cutout-panel` demo (base holes + obround slot + a hole on a folded lip).
- Also removes accidentally-committed `.claude/agent-memory-local` files (already gitignored).
- Built via a multi-agent Workflow (implement → adversarial review → fix).

### Stack
PR1–PR3 merged (#1177, #1180, #1182). Remaining: tabs/slots/form features (builds on this cutout machinery) · contour/lofted flanges · foreign-solid unfold. OCCT-WASM-first.


andymai:
Addressed the review:

**P2s fixed:**
- **Double `featureTree` walk** — `addCutout` now walks the tree once and threads it into both `worldFrames` and `developedFrame` (both gained an optional pre-computed `tree` param).
- **Undocumented `'face-0'` alias** — documented in `CutoutSpec`'s JSDoc. `'face-0'` is an established base-flat sentinel already accepted by `authorFns` (seam refs), `reliefFns`, `validateFns`, and `featureTreeFns`, so I documented it as a supported base alias (alongside `'root'`/`'base'`) rather than removing it for consistency.

**P1 disposal (buildTool wire/face on error paths)** — intentionally not changed, same convention as the rest of the package and core: the existing `buildOutline` in this same file (which your comment notes) uses the identical pattern, as do `miterFns`/`reliefFns`/`authorFns` and core `operations/`. brepjs relies on the `FinalizationRegistry` safety net for handles not created via `createHandle`/`createKernelHandle` (the only ones the pattern rule covers). Disposing only here would make `cutoutFns` inconsistent with the rest of the codebase.

PR #1186, opened 2026-06-03
feat(sheetmetal): tabs, tab-and-slot joints, louvers, dimples/embosses

andymai:
## Phase 2 · PR5 of a stacked sequence

Adds form features across the additive / assembly / formed axis, on top of the cutout machinery from PR4.

### What's new
- **Tabs** (`addTab`): a rectangular protrusion fused onto a region's edge via its world frame — adds material (volume rises) and *extends* the developed outer outline (the inverse of a cutout).
- **Tab-and-slot joints** (`tabAndSlot`): a tab plus a mating slot cutout sized `tab + clearance`, positioned so the joint self-locates when folded. The slot opening is **numerically guaranteed** larger than the tab cross-section (the test that matters for a part that actually assembles).
- **Louvers** (`louver`): an open three-side flap cut + hinge fold line — the fabricable flat representation a brake/laser operator needs — with a valid raised-flap solid.
- **Dimples / embosses** (`emboss`): raised/recessed round forms (material-neutral footprint markers in the flat).

### Design
- `TabSpec`/`TabFeature`, `FormSpec`/`FormFeature`; `FlatPattern` gains `formCuts`/`formMarkers`/`formHinges`; `FoldRegion`/`FlatInput` carry `tabs`/`forms`; `dxfFns` gains a `FORM` layer (open polylines for louver cuts).
- Forms are recorded, valid single solids. True forming is impractical with a CSG-only kernel, so the 3D form is simplified (documented) — the **flat pattern is the fabrication-critical output**, and it's correct.
- Guards: out-of-bounds, sever (`height ≥ thickness` rejected for dimples), unknown-region; `attachTabs`/`attachForms` fail loud (`TAB_REGION_UNMAPPED`/`FORM_REGION_UNMAPPED`) rather than silently dropping a feature.

### Round-trip
- Tabs round-trip through the **real outline oracle** (`partToFlatInput → fold`, Δvol ~1e-12) — review disproved an initial assumption that protrusions can't be recovered (the parser reads base/flange rects from bend lines, off the tab span). Forms are material-neutral and ride as recorded features.

### Verification
- typecheck / lint / build clean; **153 tests** (130 prior + 23 new) green against occt-wasm; snapshot adds `tab-and-slot-box` + `louvered-panel` demos (both round-trip).
- Built via a multi-agent Workflow (implement → adversarial review → fix); review caught a silent tab-round-trip gap (no `attachTabs`) and a louver flat-pattern that emitted a *closed* cut loop (would drop the flap out at fabrication) — both fixed.

### Stack
PR1–PR4 merged (#1177, #1180, #1182, #1184). Remaining: contour/lofted flanges · foreign-solid unfold. OCCT-WASM-first.

### DIFF
--- a/packages/brepjs-sheetmetal/README.md
+++ b/packages/brepjs-sheetmetal/README.md
@@ -30,7 +30,9 @@ edge that the unfold leaves uncut as a `SEAM_CUT`, flattening into a valid conne
 | Fold            | `FlatInput` region-tree → 3D part (inverse of unfold); round-trips `unfold(fold)`         |
 | Reliefs         | bend reliefs (slots at partial-flange bend-line ends), corner reliefs (notch at a corner) |
 | Cutouts         | holes / slots (rect + obround) / polygons punched on the base or a folded flange          |
-| Miter / outputs | auto corner-miter, multi-layer DXF, JSON bend report, manufacturability warnings          |
+| Tabs / joints   | additive edge tabs, self-fixturing tab-and-slot joints (tab + matching mating slot)       |
+| Form features   | louvers (3-side cut + formed flap), embosses / dimples (round raised / recessed forms)    |
+| Miter / outputs | auto corner-miter, multi-layer DXF (incl. FORM layer), JSON bend report, warnings         |
 | API             | functional `*Fns` → short-named `api.ts` → fluent `sheetMetal()` facade                   |
 
 `fold(input: FlatInput)` folds a flat pattern back up into a 3D part — the inverse of `unfold`. A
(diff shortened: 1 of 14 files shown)


andymai:
Addressed the review:

**P2s fixed:**
- **`wire` shadow** (unfoldFns) — renamed the local to `markerWire`.
- **Reused `FORM_LOOP_TOO_SMALL`** — `openPathWire` now uses a distinct `FORM_PATH_TOO_SHORT`.
- **`tabAndSlot` slot always 0°** — `SlotPlacement` gains optional `angleDeg`, threaded into the mating slot so the joint can be oriented to match the tab (+regression test).
- **Undocumented miter suppression with tabs** — `unfold` now emits a `MITER_NOT_DEVELOPED` warning (in the Ok payload) when a tab causes the developed corner-miter chamfer to be omitted; the 3D miter cut is unaffected.

**P1 disposal (3×)** — intentionally not changed, same convention as the rest of the package and core (merged `miterFns`/`reliefFns`/`cutoutFns`/`authorFns` and core `operations/` build the same fuse/cut/translate chains without manual disposal). brepjs relies on the `FinalizationRegistry` safety net for handles not created via `createHandle`/`createKernelHandle` (the only ones the pattern rule covers); disposing only in `tabFns`/`formFns` would make them inconsistent with the codebase.

PR #1190, opened 2026-06-03
feat(sheetmetal): foreign-solid import & unfold via numerical cylinder fitting

andymai:
## Phase 2 · PR7 of a stacked sequence (final)

Unfolds an **arbitrary imported sheet-metal solid** that has no feature tree, by detecting its bends from the B-rep — inverting the whole domain's "author then unfold" premise. **Pure TypeScript on occt-wasm; no kernel or WASM changes.**

### The key finding
`getSurfaceCylinderData` is `notImplemented()` on occt-wasm, and porting it would need a WASM rebuild (occt-wasm exposes a purpose-built C++ API, not raw OCCT classes). That's avoided entirely: **`fitCylinder` recovers the bend geometry numerically in TS** from the existing public surface queries — sampling `pointOnSurface`/`normalAt` across a face, taking the axis as the common perpendicular of the sampled normals and the radius from a Kåsa circle fit. This recovers **axis + radius** (more than the kernel method's `{radius, isDirect}`) with zero kernel changes.

### What's new
- `unfoldForeignSolid(solid, { kFactor? })`: classifies faces via `getSurfaceType`, pairs panel faces into flats (mid-surface; thickness = the common face-pair gap) and inner/outer cylinder faces into bends, builds the bend graph via `sharedEdges`, spanning-trees it (non-tree bends → `SEAM_CUT`), and develops each bend at the neutral axis (K=0.5 mid-surface by default, overridable for a known material).
- `fitCylinder(face)` (recovers a known cylinder's radius to 6 places, axis to ~1e-6; rejects non-cylinders by residual). `unfoldSolid` (api), `fromSolid(solid).unfold()` (facade), index exports, harness `foreign-unfold` demo.

### Honesty
Supported class: roughly-uniform-thickness solids with planar panels + cylindrical bends. Outside it — cones/splines (`UNSUPPORTED_FACE`), fit failures, non-sheet topology — it **warns rather than emitting a confidently-wrong flat** (verified: a fused sphere flags its faces; a bare sphere returns a clear `Err`).

### The oracle (non-circular)
`author(part) → part.solid → unfoldForeignSolid(solid)` reproduces `unfold(part)`'s developed area + bend count for single-bend / L-bracket / U-channel, **reading only the solid geometry** — `foreignUnfoldFns` imports no feature-tree types and the test passes only `part.solid`. Harness L-bracket: authored 3354.16 mm² vs detected 3354.16 mm², Δ = 6.78e-8.

### Verification
- typecheck / lint / build clean; **174 tests** (161 prior + 13 new) green across all kernels; snapshot foreign-unfold demo Δ area 6.78e-8 mm², 0 warnings. **0 `src/kernel/` changes.**
- Built via a multi-agent Workflow (implement → adversarial review → fix); review confirmed the fit is real (not faked) and the oracle is genuinely non-circular.

### Stack — complete
PR1 multi-bend · PR2 flat→fold + round-trip oracle · PR3 reliefs · PR4 cutouts · PR5 tabs/forms · PR6 contour/lofted · **PR7 foreign-solid unfold**. All OCCT-WASM-first.

### DIFF
--- a/packages/brepjs-sheetmetal/README.md
+++ b/packages/brepjs-sheetmetal/README.md
@@ -176,11 +176,46 @@ Both flanges have **non-rectilinear** developments, so — like miters and tabs 
 validity rather than a fold round-trip. The contour flange's developed strip still joins the rectilinear
 outline union; the lofted flange's triangulated blank is a separate developed boundary.
 
+## Foreign-solid import & unfold
+
+`unfoldSolid(solid, { kFactor? })` (fluent: `fromSolid(solid).unfold()`) flattens an **arbitrary imported
+sheet-metal solid that has no feature tree** by detecting its geometry numerically — the reverse of the
+authored path. It reads only the B-rep:
+
+- **Classify faces** by surface type: planar faces are panel faces, cylindrical faces are bend faces. Any
+  other surface type (cone, sphere, spline, …) is reported `UNSUPPORTED_FACE` and skipped (the unfold never
+  silently invents a flat for an unrecognised face).
+- **Pair flats**: the two large parallel planar faces of a panel, a sheet thickness apart, are the
+  mid-surface flat. Thickness is detected as the smallest opposed-planar-pair gap.
+- **Fit bends**: `fitCylinder(face)` recovers a bend's axis + radius + swept angle from sampled points and
+  normals. The axis direction is the common perpendicular of the sampled normals (every cylinder normal is
+  ⟂ the axis), recovered as the sign-aligned average of cross products of non-parallel normal pairs; the
+  axis line and radius come from a least-squares circle fit of the points projected onto the plane ⟂ axis;
+  the radius is the mean point-to-axis distance. A fit whose residual exceeds tolerance is rejected (warned).
+  Recovery is high-precision — radius and axis of a known `cylinder()` primitive are recovered to ~1e-6.
+- **Build the bend graph** (flats = nodes; a bend connects the two flats it shares edges with), take a
+  spanning tree from a root flat, and turn non-tree bends into seam cuts (`SEAM_CUT`, warned).
+- **Unfold** by walking the tree, replacing each cylindrical region with a developed strip of length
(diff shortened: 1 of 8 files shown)


andymai:
Addressed all three findings in 56eb0ad2:

1. **P1 — silent zero area** — `measureArea` failure now falls back to the other face, then a bounding-box estimate, and pushes a `DETECTION_INCOMPLETE` warning, so `developedArea` is never silently wrong.
2. **P2 — overloaded `INVALID_SOLID`** — split into dedicated codes for programmatic dispatch: `UNSUPPORTED_FACE` (unknown surface type / failed cylinder fit) and `DETECTION_INCOMPLETE` (non-uniform thickness, dropped panel, bend not joining exactly two flats). `INVALID_SOLID` now means only a genuinely invalid B-rep. Test asserts the `code` (not the message).
3. **P2 — non-uniform thickness silently drops panels** — a dropped *panel-sized* planar face now warns `DETECTION_INCOMPLETE`. Gated on the face's smaller in-plane dimension > 4·T so bend-edge rim bands stay quiet (area alone false-flags long rims — verified the clean-bracket demo stays at 0 warnings).

174 tests green; no `src/kernel` changes.

andymai:
The latest Greptile pass re-surfaced the same three findings, but all are already resolved in 56eb0ad2 (it re-emitted against the cumulative diff, as it has on earlier PRs in this stack):
- **silent zero area** — fixed: `measureArea` failure falls back to the other face → bounding-box estimate + `DETECTION_INCOMPLETE` warning (foreignUnfoldFns.ts ~451-470).
- **`INVALID_SOLID` overload** — fixed: now `UNSUPPORTED_FACE` / `DETECTION_INCOMPLETE`; only a genuinely invalid B-rep keeps `INVALID_SOLID` (1 remaining usage).
- **non-uniform thickness** — fixed: a dropped panel-sized face warns `DETECTION_INCOMPLETE` (gated on smaller in-plane dim > 4·T so rims stay quiet).

CI green, no new concrete defects — merging at the no-concrete-defect bar.

PR #1193, opened 2026-06-03
feat(sheetmetal): bend tables with allowance/deduction lookup + interpolation

andymai:
## Phase 3 · PR1 of 4 (bend tables · hems/jogs · nesting)

Replaces the K-factor formula with real shop **bend tables** when a part references one. The `BendRule` schema already reserved `allowance`/`deduction`/`bendTableRef` in Phase 2 — this wires them through the single resolution point.

### What's new
- **`bendTableFns.ts`**: `BendTable { id; kind: 'allowance' | 'deduction'; rows }`, a registry, and **`resolveBendAllowance`** — the single source of truth. Precedence: `bendTableRef` (interpolated table) → explicit `rule.allowance` → K-factor.
- **Interpolation**: linear-in-angle + bilinear-in-(thickness, radius); sparse one-radius-per-gauge grids resolve missing corners by nearest-radius substitution; genuinely out-of-range queries **clamp (no extrapolation)** and fire a warning via an optional sink.
- **Deduction↔allowance**: `OSSB = (R+T)·tan(θ/2)`, `BD = 2·OSSB − BA` (hand-verified for a 90° steel case; a deduction table yields the same BA as the equivalent allowance table).
- **Wiring**: `allowanceFns.developedLength`/`bendAllowance` delegate to `resolveBendAllowance`, so a `bendTableRef` part develops per the table *everywhere* (unfold/report/contour/foreign/fold/relief). Non-table parts are byte-for-byte unchanged.
- Starter air-bend tables (`steel-airbend`, `aluminum-airbend`), cited. `api`/`index` surfaces; `unfold` threads a warning sink; harness `bend-table` demo (Δ 0.223 mm vs K-factor).

### Verification
- typecheck / lint / build clean; **192 tests** (174 prior + 18 new) green; snapshot exit 0.
- Tests assert interpolation against hand values (75° midpoint = 2.5, bilinear = 5.25), the deduction-conversion 90° case, table-vs-K-factor divergence end-to-end through `unfold`, out-of-range clamp+warn, and **no spurious warning on in-range sparse-grid queries**.
- Built via a Workflow (implement → adversarial review → fix) in an isolated git worktree. Review hand-verified the math; I additionally fixed a spurious-clamp-warning on diagonal-table interior queries.

### Known minor limitation
The clamp-warning sink is currently threaded only through `unfold`; contour/fold/relief/foreign call `developedLength` without it, so a *table clamp* reached via those paths computes the (safe, clamped) value without surfacing a warning. Math is unaffected — diagnostic only. Can thread it through in a follow-up if desired.

### Stack
Phase 2 (7 PRs) merged. Phase 3 remaining: hems & jogs · bbox nesting · true-shape (NFP) nesting.

### DIFF
--- a/packages/brepjs-sheetmetal/README.md
+++ b/packages/brepjs-sheetmetal/README.md
@@ -11,9 +11,10 @@ Pipeline: **author part → auto-miter corners → fold to 3D → unfold to flat
 ## Scope
 
 Straight (cylindrical) bends, authoring our own parts (not unfolding foreign solids). The bend model is
-K-factor based with an extensible schema, so it can grow toward bend-tables without a rewrite.
+K-factor based by default, with optional **shop bend tables** (see below) for parts that develop per a
+press-brake's measured allowance/deduction values.
 
-- Bend allowance: `BA = (π/180)·|angle|·(R + K·T)`
+- Bend allowance: `BA = (π/180)·|angle|·(R + K·T)`, or interpolated from a referenced bend table.
 - Defaults: units are mm; K-factor `0.44`; inner radius `= thickness`.
 
 Authoring supports arbitrary bend trees: flanges off any of the four base edges, **chained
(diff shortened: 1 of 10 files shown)


andymai:
Both fixed in the latest commit:
- **P1 MIN_RADIUS overload** — added a dedicated `TABLE_CLAMP` warning code for table range-clamps, so it no longer collides with the `MIN_RADIUS` (R < T) condition that validateFns/unfoldFns emit and tests filter on. Clamp test asserts `code === 'TABLE_CLAMP'`.
- **P2 dead-code guard** — removed the unreachable `REGISTER_FAILED` branch (Map.get right after Map.set can't be undefined).

192 tests green.

PR #1195, opened 2026-06-03
feat(sheetmetal): hems (closed/open/teardrop/rolled) + jogs

andymai:
## Phase 3 · PR2 of 4 (bend tables ✓ · **hems/jogs** · nesting)

Two ergonomic edge features built on the contour-flange segment chainer (PR6) with table-aware developed lengths (PR8).

### Hems — `hem(part, { region, side, type, length?, radius?, gap?, rule? })`
Folds the edge back as a curl + return. Four types:
- **closed** — 180°, inner radius inflated by a documented `CLOSED_HEM_HAIR` so the near-zero-gap fold doesn't self-intersect on fuse;
- **open** — 180° at the gap radius;
- **teardrop** — ~210°;
- **rolled** — ~270°, no return.

The curl is chained as ≤120° sub-arcs (the bend-patch wedge degenerates ≥180°); the full-curl allowance is resolved **once** via `developedLength` and apportioned across sub-arcs, so a sparse bend table can't double-count (a real bug caught in review — the K-factor path's linearity had masked it).

### Jogs — `jog(part, { region, side, position, offsetHeight, angle?, … })`
Two opposite bends + a step run, sized `stepRun = (offsetHeight − (T+2r)(1−cosθ))/sinθ` so the run-out face sits **exactly** `offsetHeight` above the base — measured (not assumed) across all four kernels to 4 decimals. Rejects offsets the bends alone overshoot.

### Integration
- Both reuse the exported contour chainer and are **region-general** (off the base or any flange via `worldFrames → regionEdge`); record `hem::`/`jog::` sub-bends (skipped in the spanning tree like `contour::`); develop via `allowanceFns.developedLength` so **bend tables apply** (tied-in test).
- `unfold` develops strips + bend lines; `api`/`facade`/`index` surfaces; harness `hemmed-panel` + `jogged-bracket` demos.

### Verification
- typecheck / lint / build clean; **205 tests** (192 prior + 13 new) green; snapshot exit 0.
- Reviewer independently verified jog offset exactness across kernels and all four hem types as valid single solids (incl. adversarial tight/sub-thickness cases).
- Built via a Workflow (implement → adversarial review → fix) in an isolated git worktree.

### Stack
Phase 2 (7 PRs) + PR8 (bend tables) merged. Remaining: bbox nesting · true-shape (NFP) nesting.

### DIFF
--- a/packages/brepjs-sheetmetal/README.md
+++ b/packages/brepjs-sheetmetal/README.md
@@ -35,6 +35,8 @@ edge that the unfold leaves uncut as a `SEAM_CUT`, flattening into a valid conne
 | Form features   | louvers (3-side cut + formed flap), embosses / dimples (round raised / recessed forms)      |
 | Contour flange  | open line/arc profile swept along a base edge (multi-bend section); EXACT development       |
 | Lofted flange   | ruled transition between two open profiles; triangulated development (exact if developable) |
+| Hems            | edge folded back ~180°+ (closed / open / teardrop / rolled); EXACT development              |
+| Jogs            | two opposite bends stepping a flat by a Z-offset (joggle); EXACT development                |
 | Bend tables     | shop allowance/deduction tables, angle-linear + thickness×radius-bilinear interp, clamp-warn |
 | Miter / outputs | auto corner-miter, multi-layer DXF (incl. FORM layer), JSON bend report, warnings           |
 | API             | functional `*Fns` → short-named `api.ts` → fluent `sheetMetal()` facade                     |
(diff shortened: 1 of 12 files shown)


andymai:
All three addressed:
- **P1 open-hem gap 2×** — `gap` is now the physical clear distance; inner radius = gap/2 (a 180° fold lands the return 2r away). Docstring updated; open-hem developed-length test uses gap/2.
- **P1 closed default radius** — `closed` now defaults the inner radius to ≈0 (just the HAIR clearance) so it actually folds flat; was defaulting to one thickness (visibly open). Regression test: closed-default folds tighter than `radius=thickness`.
- **P2 one-hem-per-(region,side,type)** — `HemSpec`/`JogSpec` gain an optional `id` (falls back to the generated key), so multiple hems/jogs can sit on the same edge at different offsets. Regression test: two open hems on one edge via ids.

207 tests green.

PR #1198, opened 2026-06-03
feat(sheetmetal): true-shape (no-fit-polygon) nesting for interlocking parts

andymai:
## Phase 3 · PR4 of 4 (bend tables ✓ · hems/jogs ✓ · bbox nesting ✓ · **true-shape nesting**)

Adds a **true-shape** nesting strategy so concave / L-shaped parts interlock — the documented follow-up to PR10's bounding-box nester.

### What's new
- **`polygonFns.ts`**: 2D polygon geometry + the overlap predicate (the correctness backbone) — two simple polygons overlap iff any edges cross *or* one contains a vertex of the other; `polygonsOverlapWithClearance` adds the spacing gap.
- **`NestOptions.strategy`** (`'bbox'` default | `'nfp'`): the `nfp` packer is bottom-left-fill with a raster scan over 0/90/180/270°, using the polygon predicate as the only feasibility check (slides parts into neighbours' notches). `placementTransform` generalized to arbitrary angles (0/90 output byte-identical). Utilization credits **true outline-polygon area** so the strategies compare directly.
- `index`/`api`/README surface it; harness `nfp-vs-bbox` demo.

### Correctness (the load-bearing concerns)
- **No overlap**: the predicate was verified by the reviewer with a **20,000-pair Monte Carlo** vs a dense interior-point ground truth — **0 false negatives**; the broad-phase reject with **200,000 pairs** — 0 misses. Plus an exhaustive polygon-level non-overlap test over all placed pairs (real transformed outlines, concave cases).
- **Genuine interlocking**: utilization credits true area (1200 < bbox 1600), yet `nfp` still beats `bbox` (0.821 vs 0.547) — only possible by physically fitting 2 parts/sheet. Harness: 8 L-parts → bbox 8 sheets/54.7% vs **nfp 4 sheets/82.1% (+50%)**.
- Within-bounds, finite/terminating raster, no infinite loop, oversized → `unplaced` + warning. `bbox` stays the default so all prior behavior is unchanged.

### Verification
- typecheck / lint / build clean; **232 tests** (218 prior + 14 new) green; snapshot exit 0.
- Built via a Workflow (implement → adversarial review → fix) in an isolated git worktree.

### Stack — Phase 3 complete
PR8 (bend tables) · PR9 (hems/jogs) · PR10 (bbox nesting) merged. This finishes the bend-tables / hems-jogs / nesting plan.

### DIFF
--- a/packages/brepjs-sheetmetal/README.md
+++ b/packages/brepjs-sheetmetal/README.md
@@ -217,35 +217,61 @@ measured jog offset, and solid validity.
 
 ## Nesting
 
-`nest(patterns, { sheet, margin?, spacing?, allowRotation? })` arranges a set of developed flat patterns onto
-stock sheets to reduce waste, returning `Result<NestResult>` (`{ sheets, unplaced, warnings }`). Each sheet
-carries its `placements` (`{ patternIndex, x, y, rotationDeg }`, where `rotationDeg` is `0` or `90`) and a
-`utilization`. `nestToDXF(result, patterns, sheetIndex)` then emits one fabrication-ready, multi-layer DXF per
-sheet, translating and rotating every placed pattern (outline + bend lines + holes + forms) onto the sheet via
-the same bounding boxes the packer used.
+`nest(patterns, { sheet, margin?, spacing?, allowRotation?, strategy? })` arranges a set of developed flat
+patterns onto stock sheets to reduce waste, returning `Result<NestResult>` (`{ sheets, unplaced, warnings }`).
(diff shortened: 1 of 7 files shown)


andymai:
Both fixed (P2, docstring-only — no behavior change):
- `transformPolygon` docstring now says rotate-then-translate (matching the impl); matters now that it's exported.
- `pointInPolygon` docstring corrected — strict ray-cast, boundary behaviour undefined (edge contact is handled separately by the segment-intersection path in the overlap predicate), removed the false EPS-slack claim.

232 tests green.

PR #1219, opened 2026-06-04
feat(brepjs-verify): eval-driven skill, hint, and reference improvements

andymai:
Ran the brepjs-verify authoring loop with **escalating model tiers** (haiku → sonnet → opus) against the published 0.3.0 skill + CLI, and fixed everything that tripped them up. Each finding below came from a model failing or struggling, verified against ground truth via the CLI.

## What the eval found → what changed

**Round 1 (haiku) — all-edge fillet:** wrote `fillet(bracket, 5)` (no edge list) because the skill never showed how to *get* edges → `FILLET_FAILED`, and the hint table went dark (the `unwrap()`-throw flattened the code to a string).
- `runPart`: recover the `BrepError` code from an `unwrap()`-thrown message so hints still fire.
- `report`: add `FILLET_FAILED` / `CHAMFER_FAILED` hints.
- `references/modifiers.md`: a **Selecting edges/faces** section (edgeFinder/faceFinder) + the all-edge-fillet warning.

**Round 2 (sonnet) — face selection:** passed first try, but had to dig into source to learn `inDirection('Z')` matches **both** ±orientations.
- Documented the gotcha + the `.when(f => getBounds(f).zMax > t)` position-discriminator in `modifiers.md` and a SKILL hard rule.

**Round 3 (opus) — revolve:** `--check` caught a **TypeScript error** — `draw().close().sketchOnPlane('XZ').face()` doesn't typecheck (`.face()` isn't on the `SketchInterface | Sketches` union). The shipped `revolved-pulley` example used that exact pattern, so it failed the skill's own `--check` step.
- Rewrote `revolved-pulley` type-clean via `polygon()` (same geometry — baseline unchanged; **eval 12/12**).
- `sketching-2d.md` + SKILL hard rule: the union/`--check` gotcha and **`revolve` angle is RADIANS** (full turn = `Math.PI*2`).

**Round 0 (me) — zero-bound assertion:** a loft base lands at `z=-1e-7`, and the `expected` comparator treats an expected `0` as exact-match-only, so a zero bound could never be asserted.
- `expected`: absolute-epsilon escape hatch in the tolerance gate (+ unit test).

## llms.txt utilization (per request)

- **Bundle** `llms.txt` + `llms-full.txt` into the package (generated at build via `scripts/copyReference.mjs`, shipped through `files`), so a CLI-only user (`npm i -D brepjs-verify`, no plugin) and the skill backstop have the full API surface offline.
- **Skill backstop**: the SKILL.md reference index now points at `llms-full.txt` for anything the curated references don't cover.
- **Audit fix**: corrected the `revolve` entry in `llms.txt`/`llms-full.txt` — it said "angle default 360 degrees"; verified empirically it's **radians**, default a **half** turn. (Heads-up: that π default + the degrees/radians inconsistency vs pattern `fullAngle` look like a brepjs-core rough edge worth a separate look.)

## Verification

- typecheck **0**, eslint **0**, **72/72** package tests (incl. new fixtures for the epsilon + code-recovery paths), deterministic **eval 12/12**.
- `revolved-pulley` rewrite produces byte-identical volume to its recorded baseline.


andymai:
Addressed the 3 P2s:
- **Code-extraction regex** — anchored to the literal `[KIND] CODE:` shape (`/\[[A-Z][A-Z0-9_]*\]\s+([A-Z][A-Z0-9_]+):/`), so a stray `]…:` elsewhere in the message can't be mistaken for the code.
- **Test specificity** — now asserts the *specific* recovered code (`FILLET_FAILED`) and its matching hint, not just that some code/hint exists.
- **copyReference** — exits non-zero when no reference files are found, so a publish can't silently ship without the bundled `reference/`.

tsc 0, eslint 0, 76/76 tests, eval 12/12.

PR #1236, opened 2026-06-05
fix(manifold): degree contract for composed transforms + implement makeCompound

andymai:
## Summary

Fixes the two stacked Manifold-kernel bugs that made gridfinity's honeycomb wall pattern render as plain walls in draft previews (the cut was a silent no-op — output triangles were exactly identical to a plain bin).

### Bug 1 — `composeTransform` treated degrees as radians
The `composeTransform` kernel boundary receives **degrees** (both OCCT adapters convert per-op internally: `occt/advancedOps.ts`, `occtWasm/transformOps.ts`), but the Manifold adapter fed `o.angle` straight to `rotationMatrix`, which expects **radians**. A composed 90° rotate became a 90-radian one, so every `transformCopy`-placed shape landed off-target. `circularPattern` (public API documents degrees) and `transformBatch` carried the same mismatch.

This is the counterpart of the #1228 rotate fix: kernel `rotate` is radians, `composeTransform`/`circularPattern`/`transformBatch` are degrees — the inconsistency between those contracts is exactly where the second adapter went wrong.

### Bug 2 — `makeCompound` was `notImplemented`
Grouping the placed hex prisms threw before any cut ran (and gridfinity's builder catches + falls back to no pattern, hence the silence). Implemented as an n-ary `Manifold.union`: mesh CSG has no non-fused grouping, and union is geometrically identical to a compound for disjoint members (multi-component manifold) while staying valid for overlapping ones — downstream booleans can't distinguish the two anyway. Member-wise sub-shape iteration is the one semantic this can't support (noted in code).

## Verification
- New `tests/parity/composedTransformParity.test.ts`: composed rotate parity, chained compose order, `circularPattern` spread, and a gridfinity-shaped hex-prism-compound wall cut — all manifold-vs-occt, all failing before the fix, green after (all 4 projections).
- Full `tests/parity/` manifold projection: 222 passed, 0 regressions.
- Fixes 3 pre-existing `transformCopy.test.ts` manifold-projection failures (4 → 1; the remaining one is the per-face evolution-hash test, a known mesh-kernel limitation).
- End-to-end through gridfinity's real `generateBin` (linked dist): honeycomb on manifold now carves — plain 14,154 tris → honeycomb 32,232, minZ = 0 (occt-wasm control: 7,996 → 26,348, minZ = 0).

### DIFF
--- a/src/kernel/manifold/builderOps.ts
+++ b/src/kernel/manifold/builderOps.ts
@@ -32,6 +32,30 @@ export function makeBuilderOps(module: ManifoldModule): KernelBuilderOps {
     );
   }
 
+  function makeCompound(shapes: KernelShape[]): KernelShape {
+    // Mesh CSG has no non-fused grouping, so an n-ary union is the closest
+    // valid semantic. Disjoint members (the common case — pattern compounds)
+    // become a multi-component manifold, geometrically identical to an OCCT
+    // compound; overlapping members merge, which downstream booleans can't
+    // distinguish anyway (cutting by a compound ≡ cutting by the union of its
+    // members). What this can't support is iterating members back out as
+    // separate sub-shapes.
+    const operands = shapes.map((s) => unwrap(s));
+    const first = operands[0];
+    if (!first) {
+      notImplemented('makeCompound (no input shapes on manifold kernel)');
+    }
+    const solid = operands.length === 1 ? first : Manifold.union(operands);
+    return wrap(
+      solid,
+      makeNode(
+        'makeCompound',
+        {},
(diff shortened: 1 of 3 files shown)


andymai:
Both findings addressed in the follow-up commit: `makeCompound` now throws a plain descriptive error for empty input (invalid input, not a missing feature), and the parity tests use runtime `ctx.skip()` so a missing manifold kernel reports as skipped rather than passed. (`it.skipIf` can't work here — `haveManifold` is only known after the async `beforeAll`, which runs post-collection.)

PR #1241, opened 2026-06-07
perf(voxel): accelerate voxelize distance pass with a BVH

andymai:
## What

PR1 of the brepjs-voxel **performance & scale** stack (BVH → narrow-band → sparse grid → Fast Sweeping reinit).

`voxelize_mesh` does two `O(triangles)` passes per voxel — the closest-point **distance** pass and the FWN **sign** pass — for an overall `O(voxels·triangles)`. This PR accelerates the **distance pass** with a hand-rolled median-split AABB BVH and branch-and-bound nearest-distance traversal, dropping it to `~O(voxels·log triangles)`.

## Results (criterion, 32³ grid + pad 2, deterministic icosphere)

| metric | 1280 tris | 5120 tris |
|---|---|---|
| **distance pass** (isolated) | **8.5×** | **15.3×** |
| end-to-end (FWN-sign-bound) | 1.05× | 1.35× |

The end-to-end ratio is intentionally modest: the FWN sign pass is **unchanged** (hierarchical FWN is a separate, deferred item), so it now dominates — which is exactly the ceiling PR2 (narrow-band + flood-fill sign) removes. The bench measures both so this is visible, not inferred.

## Design notes

- **Hand-rolled BVH, not the `bvh` crate** — that crate is ray-traversal-oriented and pulls in `nalgebra`. We need nearest-*point* branch-and-bound, and the leaf test reuses the existing exact Ericson `point_to_triangle_distance` verbatim.
- **Exact, not approximate.** The traversal returns the same linear argmin distance brute force does (squared-space pruning is monotonic in `d`; the leaf tracks the linear `best_d`), so output is **bit-exact** — a parity test asserts `<1e-6` over 4 fixtures.
- **wasm size held at 57.9 KB** (vs 52.3 KB baseline, +10.7%). A spatial-median partition avoids `select_nth_unstable_by`, whose monomorphization cost ~6.5 KB. The BVH result is exact for any partition, so this only affects tree balance — no perf regression.
- **Fully transparent**: `pkg/index.js` / `index.d.ts` are byte-identical to baseline; no API change.

## Gates

- 38 cargo tests (+7 BVH/parity), 144 vitest (voxel + lattice), `clippy -D warnings` clean.

## Follow-up (not in this PR)

`Cargo.toml` is stranded at `0.1.0` while `package.json`/release-please are at `0.2.0` because the release-please config has no `extra-files` entry for `Cargo.toml` — so `version()` reports a stale string. Worth a separate chore to add the `x-release-please-version` annotation + `extra-files: ["Cargo.toml"]`.


andymai:
Thanks for the review. Addressing both P2s:

**P2 (per-voxel `Vec` allocation in `nearest_distance`)** — fixed in f8aff2d0. Split into an allocating `nearest_distance` wrapper (kept for tests) and `nearest_distance_with(p, &mut stack)`; the `voxelize_mesh` / `distance_field_bvh` loops now feed a single grow-once buffer instead of allocating per voxel. Distance-pass bench improved slightly even on native (58.9→56.9 ms / 116.7→114.5 ms at 1280/5120 tris); the wasm allocator should benefit more. Output unchanged — bit-exact parity still holds.

**P2 (`pub mod fwn/grid/ops` widens the surface)** — respectfully declining. The crate is `publish = false` with no Rust consumers, and the *wasm* export surface is determined by `#[wasm_bindgen]`, not `pub mod` — `pkg/index.js` and `index.d.ts` are byte-identical to baseline (verified after both commits). The suggested `#[cfg_attr(target_arch = "wasm32", allow(dead_code))]` on `fwn`/`grid` would also be inaccurate: those modules are *live* on wasm32 (the bindgen entry points in `lib.rs` call into them), so they aren't dead code there. `pub` is the minimal visibility the external criterion harness needs to reach `Mesh`/`Grid`/the bench shims. Happy to revisit if a real Rust consumer ever appears.

PR #1245, opened 2026-06-07
perf(voxel): hierarchical Barnes-Hut FWN for the voxelize sign pass

andymai:
## What

PR2 of the brepjs-voxel **performance & scale** stack. After PR1 (BVH distance), the FWN **sign** pass — `mesh.is_inside(p)` per voxel, `O(voxels·triangles)` — was ~98% of `voxelize_mesh` end-to-end. This accelerates it to `~O(voxels·log triangles)` with a Barnes-Hut / Barill-2018 hierarchical winding number that **reuses the PR1 BVH** (one tree, two traversals).

Each BVH node carries a first-order point-dipole expansion (area-weighted normal sum about the area-weighted centroid, aggregated bottom-up). A query approximates a node by its dipole when `|p − center| > β·radius` (β = 2), else recurses and evaluates the **exact** solid angle at leaves.

## Results (icosphere, 32³ + pad 2)

| pass | 1280 tris | 5120 tris |
|---|---|---|
| **sign pass** (exact → hierarchical) | 29.6× | **107×** |
| **end-to-end** | 14.5× | **36.1×** (7.14 s → **0.20 s**) |

Log-scaling is visible: the sign-pass speedup grows 30→107× as N quadruples.

## Correctness — sign identical to exact FWN

- **Zero sign mismatches** vs the exact O(N) FWN oracle across every voxel of unit-cube, **holey-cube (the keystone)**, and icosphere. The keystone (holey interior stays inside, w ≈ 0.833) routes through the fast path unchanged.
- The winding **number** is approximate; **the opening criterion — not expansion accuracy — guarantees the sign**. Near-0.5 (near-surface) nodes are evaluated exactly; the dipole is used only for far nodes where w is unambiguously ~0 or ~1.
- The β = 2 dipole-only margin is **empirical, not a proven bound** (the docstring says so). Committed **adversarial guards** — high-aspect slab, far disjoint emitter, single open triangle — lock it so a future β/leaf-size change can't silently erode it.

## Notes

- API fully transparent: `pkg/index.js` / `index.d.ts` byte-identical; one scratch stack serves both traversals.
- wasm 57.8 → 60.2 KB (+2.4 KB; first-order only, no 2nd-order tensor — measured unnecessary at β = 2).
- 47 cargo tests (+9), 144 vitest, clippy clean.

Built via Workflow (design → implement → review). Both review angles checked; the one "changes-requested" flagged 5 missing expansion unit tests — verified **false** (all 5 are present in `bvh.rs`; `cargo test` = 47; the reviewer caught a mid-edit snapshot). Legitimate P3s (docstring honesty, adversarial guards, stack reuse, bench attribution) all applied.


andymai:
Good catch on the **sign_pass benchmark methodology** (P2). You're right that `sign_field_fast` calls `Bvh::build` inside the timed region while the exact baseline has no setup — so the reported 29.6×/107× are a *conservative* lower bound on the true per-query traversal gain, since production `voxelize_mesh` builds the BVH once and amortizes it across both passes (and `end_to_end` already captures that real combined cost).

I went with your "at minimum" suggestion — a clear caveat comment on both the `sign_pass` and `distance_pass` groups noting the build is inside the timed region and the ratios understate the win — rather than hoisting the build out via a prebuilt-`Bvh` shim. Doing the hoist cleanly would require making `mod bvh` public and adding traversal-on-prebuilt-BVH shims, i.e. widening the crate's Rust surface — the exact thing flagged on PR1. Since the bias is conservative (numbers only get *better* with the fix) and the production path is what `end_to_end` measures, the transparency comment is the better trade. Fixed in the latest commit.

PR #1249, opened 2026-06-07
perf(voxel): sparse tiled grid + seam-free tiled contouring

andymai:
## What

PR4 (final scale PR before Fast Sweeping) of the brepjs-voxel stack. Lifts the resolution ceiling from **O(volume) to O(surface)** memory. After PR3 narrow-band, far-field voxels are uniformly `sign·band`, so a dense grid is mostly redundant — store only near-surface tiles.

- **`src/sparse.rs` (new):** `SparseGrid` — 16³ tile blocks in an int-keyed map, lazy allocation, implicit far-sign oracle, `MAX_ACTIVE_TILES` budget, `offset_far`/`shell_far`.
- **`src/contour.rs`:** `surface_nets_mesh_sparse` — a from-scratch **global-cell-coord-keyed** dual Surface Nets. Each surface cell is owned by exactly one tile; cross-tile quads resolve through the global vertex map. **Seam-free by construction**, reproducing the dense contour exactly (replicates fast-surface-nets 0.2's estimator).
- **`src/ops.rs`:** `voxelize_mesh_sparse` — AABB tile activation + one-FWN-per-absent-tile far sign; bit-identical to `voxelize_mesh_banded` on active cells.
- **`src/lib.rs`:** `Pipeline` router (`DENSE_THRESHOLD=4M`). repair/offset/shell pick dense (small grids — the 144 vitest, unchanged) vs sparse (large). boolean/lattice stay dense (documented; sparse two-grid combine is future work). **No wasm-bindgen signature change.**

## Seam-freedom is gated, not asserted

The hard gate: `sparse-contour == dense-contour`, **canonicalized bit-exact** (`f32::to_bits`), on cube/sphere/holey/slab across resolutions spanning many tiles. A crack, duplicate vertex, or missing cross-tile triangle all fail it. The dense path is retained as the oracle.

## Scaling (the payoff)

| res | tiles allocated | note |
|---|---|---|
| 128 | 516 / 729 | O(area), not O(volume) |
| **408** (icosphere) | 5953 / 17576 | **dense REFUSES** (70.4M > MAX_VOXELS); sparse completes a **909k-triangle watertight** mesh from 24.4M band voxels |

A cheap always-on test proves *success above the dense cap* (small sphere in large bounds: dense errors, sparse finishes in ~2s).

## Review fixes (both reviewers requested changes — all addressed)

- **P1** — sparse offset/shell far-field math had no parity coverage (only runs on the >4M sparse route). Added a cargo parity test vs the dense oracle: **passes bit-exact** (coverage gap, not a bug).
- **P2** — centroid used reciprocal-multiply (`sum·(1/n)`); `s/n ≠ s·(1/n)` in f32 by 1 ULP for ~18% of values (count=3 common on spheres). Now true division → bit-identical to upstream on wasm.
- **P2** — the entire Rust correctness proof ran *nowhere* in CI. Added a `voxel-wasm-rust` job (cargo test + clippy). (No rebuild-and-diff of `pkg/` — `wasm-opt` output is version-sensitive and would flake.)
- **P3** — loud `debug_assert` in `maybe_quad` (a missing neighbor = seam crack); cheap always-on above-cap success test; near-surface coverage.
- Deferred (documented): routing boolean/lattice through sparse; a wasm-target dense-vs-sparse vitest (native parity + IEEE-deterministic f32 cover it).

## Notes

- wasm **60.3 → 88 KB** (+27.6 KB): a genuine second pipeline (sparse voxelizer + from-scratch tiled Surface Nets chosen for *provable* seam-freedom over crate-reuse + welding). Within the 200 KB budget.
- 148 vitest, 68 cargo tests. API transparent.


andymai:
Good catch on the **u32 tile-key overflow** — fixed in the latest commit. `SparseGrid::new` now returns `Result` and refuses with `GridError::TooLarge` (via `checked_mul`, which also catches the `ntiles` product wrap on wasm32 where `usize` is 32-bit), **before any allocation** — mirroring the dense `MAX_VOXELS` cap. That converts the silent block-aliasing corruption into a clean refusal. It's reachable exactly as you described: the sparse path's `MAX_ACTIVE_TILES` budget caps *active* tiles, not total, so a high-res request with a small surface band could slip through. Added an allocation-free test pinning the refusal.

PR #1275, opened 2026-06-13
feat(viewer): shared ViewerControls toolbar; interactive verify --serve

andymai:
## What

Turns the bare, read-only `brepjs-verify --serve` viewer into a lightweight inspector, by **extracting a reusable toolbar into the shared `brepjs-viewer`** rather than rebuilding it. This is **PR A of Phase 1** of a viewer overhaul (inspired by `earthtojake/text-to-cad`'s `cad-viewer`).

### `brepjs-viewer` (shared lib)
- New **`ViewerControls`** — store-agnostic, fully-controlled toolbar: view-mode (solid/wire/x-ray), edges, grid, turntable, view presets (iso/front/top/right), fit, screenshot. Each group renders only when its handler is supplied; self-contained inline styles so it works without Tailwind (`className` to restyle).
- **`ViewerCanvas`** now accepts `autoRotate`, `gridVisible`, and a `fitSignal` (bump to re-frame). Flips `frameloop` to `always` while spinning so the turntable advances, `demand` otherwise.

### `brepjs-verify` (consumer)
- The standalone viewer wires `ViewerControls` with local React state + an in-browser PNG download, and renders edges.
- The agent **snapshot pipeline loads the page with `ui=0`** to suppress the toolbar — captured PNGs stay pixel-clean (verified below).

## Verification
- `brepjs-viewer`: typecheck + lint + build green.
- `brepjs-verify`: typecheck + lint + **78 tests** + viewer build green.
- Snapshot path → chrome-free PNG (ui=0); serve path → full toolbar (both screenshot-confirmed locally).

## Scope / follow-ups
- Projection (ortho/persp) toggle deferred — it needs the camera-bridge work and is the riskiest bit.
- **PR B (next):** refactor the playground `ViewerToolbar` to consume this same `ViewerControls` (pure refactor, no behavior change).
- Later phases: measurements/info panel, face-pick inspection, section plane, snapshot dimension overlay.

### DIFF
--- a/packages/brepjs-verify/README.md
+++ b/packages/brepjs-verify/README.md
@@ -48,7 +48,7 @@ npx -y brepjs-verify part.brep.ts --snapshot shots/                     # iso/fr
 npx -y brepjs-verify part.brep.ts --serve                               # clickable preview link (renders the real STEP)
 ```
 
-`--snapshot`/`--serve` use the bundled viewer (shipped under `viewer/dist`, including the OCCT WASM). The viewer is read-only display + screenshot in v1.
+`--snapshot`/`--serve` use the bundled viewer (shipped under `viewer/dist`, including the OCCT WASM). The `--serve` link is interactive: a toolbar offers view presets + fit, solid/wireframe/x-ray modes, edge/grid toggles, a turntable, and an in-browser PNG screenshot. `--snapshot` loads the same page with `ui=0` to suppress the toolbar, so captured PNGs stay chrome-free.
 
 ## CLI reference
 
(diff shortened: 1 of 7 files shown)


andymai:
Both P2s addressed in 25ceb67d:
- `aria-pressed` now omitted on the one-shot Fit/Snap buttons (kept on toggles only).
- `onFirstFrame` held in a ref and removed from the framing effect deps, so an inline callback can't snap the camera back mid-orbit. Verified the snapshot pipeline still flips `__ready` and emits all four PNGs.

PR #1277, opened 2026-06-13
feat(viewer): measurements info panel in verify --serve

andymai:
## What

**Phase 2** of the viewer overhaul. Adds a measurements/info panel to the `brepjs-verify --serve` viewer — the most verify-specific feature, since it directly answers *"is this the right part?"* with size · volume · area · validity.

### `brepjs-viewer`
- New **`ViewerInfoPanel`** — controlled, store-agnostic readout (bbox size, volume, area, triangles, validity); renders only the rows whose values are present; self-contained inline styles.
- New pure **`meshSize(data)`** helper — bbox extents from mesh positions.

### `brepjs-verify`
- The kernel worker now measures **volume / area / validity** on the imported shape (`measureVolume`/`measureArea`/`isValid`), best-effort: non-positive volume and failed ops are omitted, validity always resolves. Measurements ship alongside the mesh.
- The standalone viewer renders the panel bottom-left, gated behind the same `ui` flag as the toolbar — **agent snapshots stay chrome-free**.

## Verification
- `brepjs-viewer` + `brepjs-verify`: typecheck + lint + build green; **80 tests** pass (+2 new measurement tests).
- Screenshot-confirmed on a 10×10×10 box: Size `10 × 10 × 10`, Volume `1,000`, Area `600`, Triangles `12`, Validity `valid` — all exact.

## Notes
- Bbox size is mesh-derived (within tessellation tolerance) — a visual aid, not a replacement for the CLI verify report's exact kernel bbox.

## Next phases
3 — face-pick inspection · 4 — section plane · 5 — snapshot dimension overlay.

### DIFF
--- a/packages/brepjs-verify/viewer/main.tsx
+++ b/packages/brepjs-verify/viewer/main.tsx
@@ -3,8 +3,10 @@ import { createRoot } from 'react-dom/client';
 import {
   ViewerCanvas,
   ViewerControls,
+  ViewerInfoPanel,
   Renderer,
   EdgeRenderer,
+  meshSize,
   type ViewMode,
   type ViewName,
 } from 'brepjs-viewer';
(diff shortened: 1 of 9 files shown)


andymai:
Addressed in 5573b7df — `area` now uses the same `> 0` guard as `volume`, so a buggy or unsupported-shape area can't reach the panel.

PR #1278, opened 2026-06-13
feat(viewer): click-to-inspect face picking in verify --serve

andymai:
## What

**Phase 3** of the viewer overhaul. Click a face in the `--serve` viewer to inspect it — highlighted in the scene, with its surface type, area, and model-space normal in a panel.

### `brepjs-viewer`
- New **`ViewerSelectionPanel`** — controlled readout for a picked `FaceInfo` (surface type, area, normal) with a clear button; renders nothing when no face is selected.

### `brepjs-verify`
- The kernel worker now collects per-face metadata — `getSurfaceType` / `measureArea` / `normalAt`, keyed by `getHashCode` so it matches the mesh's `faceGroup` faceIds — with per-face try/catch (a degenerate face can't wipe the rest). Mirrors `apps/playground`'s `collectFaceInfos`.
- The viewer wires `Renderer.onFacePick`/`onFaceHover`, highlights the picked face via `SelectionHighlight`, and shows the selection panel top-left.
- Collection is gated on an `inspect` flag threaded through `useModel` → worker, so **headless snapshots (`ui=0`) skip the extra per-face kernel work** and capture as fast as before.

## Verification
- typecheck + lint + builds green; **83 tests** pass (+3 new: default-skip, collect, per-face-throw resilience).
- Diagnostic confirmed mesh `faceId` ≡ `getHashCode` and correct axis-aligned normals on a box.
- Screenshot-confirmed: clicking the cube's +Z face highlights it teal and the panel reads `Planar face · Area 100 · Normal 0, 0, 1` (model coordinates, matching the STEP/report frame).

## Next phases
4 — section/clipping plane · 5 — snapshot dimension overlay.

### DIFF
--- a/packages/brepjs-verify/viewer/main.tsx
+++ b/packages/brepjs-verify/viewer/main.tsx
@@ -4,9 +4,12 @@ import {
   ViewerCanvas,
   ViewerControls,
   ViewerInfoPanel,
+  ViewerSelectionPanel,
   Renderer,
   EdgeRenderer,
+  SelectionHighlight,
   meshSize,
+  type FaceInfo,
   type ViewMode,
   type ViewName,
 } from 'brepjs-viewer';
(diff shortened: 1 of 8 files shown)


andymai:
Addressed both in da5b8ee4:

- **L96 (silent catch):** done — `console.warn` now fires when `getFaces` throws or a single face's metadata extraction fails, matching the playground worker so a missing pickable face is diagnosable.
- **L98 (face-handle disposal):** intentionally not changed — this is a false positive for this API. `getFaces` returns **shape-owned cached handles** (`topologyQueryFns.ts` stores them in a `topoCache` WeakMap keyed by `shape.wrapped` and returns the same array on repeat calls). They aren't fresh per-call allocations, so wrapping them in a per-iteration `DisposalScope` would dispose cached handles still owned by the shape — a double-free / use-after-free on the next `getFaces`. The playground deliberately doesn't dispose them either, and the repo's own `check:patterns` disposal rule passes on this loop. Their lifetime is tied to the shape.

PR #1279, opened 2026-06-13
feat(viewer): section/clipping plane in verify --serve

andymai:
## What

**Phase 4** of the viewer overhaul. Adds a movable **section plane** to the `--serve` viewer so internal features (bores, shelled cavities, pockets) can be inspected.

### `brepjs-viewer`
- New helpers **`sectionPlane(axis, position, flip)`** (model-space `THREE.Plane`) and **`meshBounds(data)`**.
- `Renderer` and `EdgeRenderer` gain an optional **`clippingPlanes`** prop wired to their materials.
- `ViewerCanvas` enables **local clipping** (`gl.localClippingEnabled`), so only the model is cut — the grid stays intact.
- New **`ViewerSectionControls`** — controlled bar: enable toggle, axis (X/Y/Z), position slider, flip.

### `brepjs-verify`
- The viewer owns the section state, derives the slider range from the model bbox, and feeds the plane to the renderers. **Off by default, gated behind the `ui` flag** — headless snapshots are unaffected.

## Verification
- typecheck + lint + builds green; **87 tests** pass (verify 83 + viewer-lib geometry: new `meshBounds`/`meshSize`/`sectionPlane` cases).
- Screenshot-confirmed on the `hollow-enclosure` example: a mid-X section reveals the internal cavity and wall thickness while the grid is uncut; the section bar shows Section·X active with the slider centered.

## Notes
- The cut is open (no solid cap face) — standard for a quick clip; a stencil cap would be a future enhancement.

## Next
Phase 5 — dimension overlay burned into agent snapshots (the last phase).

### DIFF
--- a/packages/brepjs-verify/viewer/main.tsx
+++ b/packages/brepjs-verify/viewer/main.tsx
@@ -1,15 +1,19 @@
-import { StrictMode, useCallback, useEffect, useState } from 'react';
+import { StrictMode, useCallback, useEffect, useMemo, useState } from 'react';
 import { createRoot } from 'react-dom/client';
 import {
   ViewerCanvas,
   ViewerControls,
   ViewerInfoPanel,
   ViewerSelectionPanel,
+  ViewerSectionControls,
   Renderer,
   EdgeRenderer,
   SelectionHighlight,
   meshSize,
+  meshBounds,
+  sectionPlane,
   type FaceInfo,
+  type SectionAxis,
   type ViewMode,
   type ViewName,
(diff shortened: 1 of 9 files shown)


andymai:
Both addressed in 956ec97a:
- **L81:** `ViewerCanvas` now resets `gl.localClippingEnabled = false` on unmount, matching the file's other teardown pattern.
- **L61:** the position slider is disabled + dimmed when the axis has zero extent (`min === max`), so a stuck handle reads as intentional.

PR #1281, opened 2026-06-13
feat(viewer): orthographic/perspective projection toggle

andymai:
## What

Viewer-overhaul **follow-up B**. Adds an ortho/persp projection toggle to the `--serve` viewer.

### `brepjs-viewer`
- `ViewerCanvas` gains a **`projection`** prop. When `'orthographic'`, it mounts a `makeDefault` `OrthographicCamera` (starting at the perspective camera's position for a continuous swap) and the framing effect fits via **zoom** instead of distance. Drei restores the perspective camera on switch-back.
- `ViewerControls` gains an optional **projection toggle** button (`Persp`/`Ortho`), rendered only when `onToggleProjection` is supplied.
- Exports a `Projection` type.

### `brepjs-verify`
- The viewer wires projection state + the toggle button, and passes `projection` to `ViewerCanvas`.

## Verification
- typecheck + lint + builds green; 83 verify tests + 4 viewer-lib tests pass.
- Screenshot-confirmed: toggling to Ortho renders a box with parallel edges (true parallelogram top face, no foreshortening); the `Ortho` button shows active.

## Context
This is 1 of 2 viewer-overhaul follow-ups. The next PR refactors the playground's `ViewerToolbar` to consume this shared `ViewerControls` (now that it supports projection) — which also reconciles the two preset vocabularies (`right`↔`side`, `iso`↔`isometric`).

### DIFF
--- a/packages/brepjs-verify/viewer/main.tsx
+++ b/packages/brepjs-verify/viewer/main.tsx
@@ -13,6 +13,7 @@ import {
   meshBounds,
   sectionPlane,
   type FaceInfo,
+  type Projection,
   type SectionAxis,
   type ViewMode,
   type ViewName,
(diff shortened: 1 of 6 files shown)


andymai:
Both addressed in cb88c7a7:
- **L18:** moved the `Projection` union into `types.ts` (next to `ViewMode`/`ViewName`); `ViewerCanvas` and `ViewerControls` both import it from there, so there's a single definition.
- **L82:** corrected the `OrthoCamera` comment — the captured position only seeds the first frame (avoids an origin flash before `Framing`'s passive effect runs); `Framing` then sets the final position and zoom-fit. Behavior is unchanged and the ortho framing was screenshot-verified.

PR #1289, opened 2026-06-13
fix(kernel): thread STL tolerance/angularTolerance through brepkit + occt-wasm

### LINKED ISSUE #1269: BrepkitAdapter.exportSTL ignores tolerance/angularTolerance arguments
`BrepkitAdapter.exportSTL` drops its `tolerance` and `angularTolerance` arguments and always exports with `DEFAULT_DEFLECTION = 0.01`. Callers passing a custom tolerance silently get the default.

Currently no practical effect for the gridfinity tool (its `EXPORT_TOLERANCE` equals 0.01), but the API contract is broken: the arguments should be threaded through to the underlying tessellation/export call.

andymai:
## Summary

Fixes #1269. `BrepkitAdapter.exportSTL` silently dropped its `tolerance` and `angularTolerance` arguments and always exported at the fixed `DEFAULT_DEFLECTION = 0.01`, so callers passing a custom tolerance got the default.

The interface `KernelIOOps.exportSTL(shape, binary?, tolerance?, angularTolerance?)` already declared these params — this was an adapter-conformance bug.

## Changes

- **New `src/kernel/stlBuilder.ts`** — kernel-agnostic STL serializers: `buildBinarySTL` (extracted from occt-wasm) + new `buildAsciiSTL`. Facet normals derived from triangle winding.
- **brepkit `exportSTL`** now routes through `mesh()` (which threads both tolerance and angularTolerance to `tessellateSolid`) and builds STL from the triangle soup, instead of calling native `exportStl`/`exportStlAscii` with a hardcoded deflection. Adapter wiring updated to pass the two args through.
- **occt-wasm `exportSTL`** ASCII path now also builds from `mesh()` (it previously used native `exportStl`, which only takes linear deflection), so `angularTolerance` is honoured for both binary and ASCII. Removes the duplicated `buildBinarySTL` in favour of the shared one.

## Tests

- Added two regression tests in `tests/meshFns.test.ts` asserting that coarser vs finer **linear tolerance** and **angularTolerance** produce different STL facet counts — these would have caught the original bug. Pass on occt-wasm, occt, and brepkit.
- Registered a `meshFns.exportStlTolerance` manifold divergence (manifold's sphere is a fixed-segment primitive whose facet count is tolerance-independent).

## Verification

- `typecheck`, `eslint`, `check:boundaries`, `prettier --check`, `knip` — clean.
- `check:patterns` — no new violations from this change (the 21 reported pre-exist on `main` in `src/kernel/manifold/*`).
- STL/IO suites green on occt-wasm; new tests green on occt-wasm/occt/brepkit; manifold skips as registered.

### DIFF
--- a/src/kernel/brepkit/ioOps.ts
+++ b/src/kernel/brepkit/ioOps.ts
@@ -18,6 +18,8 @@ import {
   DEFAULT_DEFLECTION,
 } from './helpers.js';
 import { wasmIndex } from '@/utils/vec3.js';
+import { buildAsciiSTL, buildBinarySTL } from '@/kernel/stlBuilder.js';
+import { mesh } from './meshOps.js';
 
 export function exportSTEP(bk: BrepkitKernel, shapes: KernelShape[]): string {
   if (shapes.length === 0) return '';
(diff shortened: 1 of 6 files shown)


andymai:
Thanks @greptile-apps — both addressed in ea995b18:

- **angularTolerance default divergence:** hoisted `DEFAULT_STL_TOLERANCE` / `DEFAULT_STL_ANGULAR_TOLERANCE` into `stlBuilder.ts`; brepkit and occt-wasm now import the same constants, so their export defaults are identical and can't drift.
- **per-triangle allocation:** `facetAt` now fills a single reused module-level scratch `Facet` instead of allocating one per triangle.

Issue #1287, opened 2026-06-13
Kinematics roadmap: drivable joints + forward kinematics for assemblies


andymai:
## Context

Follow-up to #1286, which documents the gaps. That issue is gap-analysis only; this one proposes a phased roadmap to take the assembly system from static planar positioning to drivable articulated mechanisms (robot arms, linkages).

Phases are ordered by dependency: each builds on the previous, and **Phase 0 is a prerequisite for everything** — kinematics is meaningless if constraints don't compose down a chain.

## Phase 0 — Fix constraint composition *(prerequisite)*

Addresses gap A4 in #1286. The solver must position each part using the *solved* pose of the parts it references, not their original geometry.

- Carry an accumulated transform per node through `solveConstraints` and apply it to each entity's `origin`/`normal` before computing the dependent part's pose.
- Resolve constraints in topological order (fixed/base first), so a referenced part is already placed when a dependent constraint reads it.
- Flip `tests/mateFns.test.ts:424-428` from asserting the buggy z=8 to the correct z=13.

*Exit criterion:* `base→mid→top` chain composes correctly; no constraint reads pre-transform geometry.

## Phase 1 — Rotational constraints + reachable axis entity

Addresses gaps A1 and A3.

- Make `extractEntity` (`mateFns.ts:46-58`) yield an `axis` entity from a cylindrical face or a linear edge, so the solver's existing `'axis'` `SolverEntity` type becomes reachable.
- Implement `concentric` (axis–axis alignment) and `angle` in `solveConstraints`, replacing the `unsupported.push` at `solverAdapter.ts:108-109`.
- Remove the corresponding entries from `UNSUPPORTED_DOF` as each lands.

*Exit criterion:* a pin-in-hole `concentric` mate solves and converges; `angle` mate constrains relative orientation.

## Phase 2 — Joint primitives

New module `src/operations/jointFns.ts` (Layer 2), built on the assembly tree.

- `RevoluteJoint` (axis + angle range), `PrismaticJoint` (axis + travel range), with a drivable parameter (current value within range).
- A joint is sugar over Phase-1 constraints (revolute ≈ concentric + angle driver; prismatic ≈ coincident + distance driver) plus a stored DOF parameter, distinct from `AssemblyNode.rotate` (which stays a static structural transform).
- Surface through `src/topology/api.ts`.

*Exit criterion:* define a single revolute joint, set its angle, and get the correct child transform.

## Phase 3 — Forward kinematics + mechanism DOF

[...]

Issue #1286, opened 2026-06-13
Assembly system: non-plane coincident/distance mates, inverse kinematics, and URDF/DH interchange


andymai:
## Summary

This issue documents the current state and gaps in brepjs's assembly/mate system, covering both (A) the partial constraint solver and (B) the complete absence of kinematics. **Gap analysis only — no solution is proposed here.**

The assembly *tree* (`src/operations/assemblyFns.ts`) is solid: an immutable hierarchy of nodes with local `translate` / `rotate` / metadata. The gaps are all in constraint solving and motion.

## A. Constraint solver gaps

Solver: `src/kernel/solverAdapter.ts` · Mate layer: `src/operations/mateFns.ts`

**1. `concentric` and `angle` are declared but never solved.**
`solveConstraints` pushes both straight onto `unsupported` (`solverAdapter.ts:108-109`), so `solveAssembly` returns `ASSEMBLY_NOT_CONVERGED`. `concentric` reports 4 DOF unresolved, `angle` 1 DOF (`UNSUPPORTED_DOF`, `solverAdapter.ts:37-42`). `concentric` is the canonical revolute-joint mate (pin-in-hole), so its absence blocks the most basic articulated connection.

**2. `coincident` and `distance` are plane–plane only.**
Both branches fall through to `unsupported` for any non-plane entity pair (`solverAdapter.ts:85-86`, `105-106`). No cylindrical, edge, or point-based variants.

**3. `extractEntity` can only ever yield `plane` or `point` — never `axis`.**
`mateFns.ts:46-58` maps a face → `plane`, a point → `point`, and returns `null` for an edge. The solver's `'axis'` `SolverEntity` type (`solverAdapter.ts:9`) is therefore unreachable from the public mate API, and edge-based mates fail extraction with `ASSEMBLY_MATE_INVALID`.

**4. Constraints don't compose across a chain (uses original, pre-transform geometry).**
`extractEntity` always reads `faceCenter(mate.face)` from the *original* face (`mateFns.ts:48`); the solver computes each part's position from the other part's original origin, ignoring any transform a prior mate assigned it. Documented at `CLAUDE.md:109` and asserted by the test at `tests/mateFns.test.ts:424-428` (chain `base→mid→top` puts `top` at z=8 using b2's local z=5, not its solved z=10). This breaks any multi-body chain — exactly the structure of a robot.

## B. Kinematics: absent entirely

[...]

PR #1292, opened 2026-06-13
fix(assembly): compose mate constraints down a chain (kinematics Phase 0)

andymai:
## Summary

**Phase 0 of the kinematics roadmap (#1287)** — the prerequisite for everything that follows. Addresses gap **A4** documented in #1286: the constraint solver positioned each part from the referenced part's *original* geometry, ignoring any transform a prior mate assigned it, so multi-body chains mis-solved.

## What changed

`solveConstraints` (`src/kernel/solverAdapter.ts`) now resolves positioning mates in **topological order**:

- For a positioning mate, **entityA is the reference** and **entityB the dependent**.
- **Anchors** (placed at the origin): chain roots — nodes never positioned by a mate — plus explicit `fixed` nodes.
- Each `coincident`/`distance` mate solves once its reference is placed, positioning the dependent against the reference's **already-solved pose** (`solvePlanePair` applies the reference's solved translation before measuring).
- Resolution is robust to the **order mates are declared in** (rounds over a worklist, not array order).
- `concentric`/`angle` and non-plane entity pairs remain unsupported (Phase 1); non-plane pairs are still reported eagerly by entity type.

Rotation stays identity — `coincident`/`distance` produce pure translations; full quaternion composition lands in Phase 1 with `angle`/`concentric`.

## Tests

- **Flipped** the composition-bug test (`tests/mateFns.test.ts`) from the buggy `top.z = 8` to the correct **`z = 18`** (mid solved to z=10, spans 10..15; 3-unit gap → top at 18). *(The roadmap's "13" was an arithmetic slip; 18 is the face-gap-faithful value.)*
- **Added** a test proving a chain composes correctly even when mates are declared **leaf-first** (out of order).
- All 26 mateFns + 17 assemblyFns tests green; full changed-file suite (3382 tests) green via pre-push.

## Docs

- Updated the `CLAUDE.md` gotcha that documented the old "original face coordinates" limitation.

## Follow-ups (this roadmap)

- Phase 1 — rotational constraints (`concentric`/`angle`) + reachable `axis` entity
- Phase 2 — joint primitives (`jointFns.ts`)
- Phase 3 — forward kinematics + mechanism DOF

Refs #1287, #1286

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -106,7 +106,7 @@ For adding a new operation or kernel method, see `.claude/commands/`.
 
 - OCCT Emscripten returns enum objects with `.value` — extract with: `typeof val === 'number' ? val : Number(val?.value ?? val)`
 - `autoHeal` short-circuits valid shapes: `report.alreadyValid=true`, no sew/heal diagnostics run
-- Assembly solver uses original face coordinates — distance constraints don't compose across multiple mates
+- Assembly solver composes constraints down a chain: each positioning mate reads the referenced part's _solved_ pose and resolves in topological order (a `fixed` node anchors the chain). `concentric`/`angle` and non-plane entity pairs are still unsupported (report `ASSEMBLY_NOT_CONVERGED`).
 - `Uint32Array` WASM interop: always convert to regular `Array` before passing to kernel methods
 - `DisposalScope` disposes in LIFO order — register dependencies after their dependees
 - `noUncheckedIndexedAccess` is enabled — array indexing returns `T | undefined`, add bounds checks or use `!` with eslint-disable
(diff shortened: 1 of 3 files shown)


andymai:
Thanks @greptile-apps — all three P2s addressed in 65a5306a:

- **`depPos` always zero:** removed the parameter entirely. A dependent is only solved while unplaced, so its position is structurally the origin; `solvePlanePair` now returns the absolute translation. Cleaner than carrying a forward-looking param that Phase 1 will redesign anyway.
- **Unanchored branch untested:** added a two-node mutual-reference cycle test asserting `converged: false` and two `coincident(unanchored)` entries.
- **Non-plane dependent cascade:** documented at the non-plane handling that a node left unplaced by an unsupported pair cascades its downstream plane-plane mates to `(unanchored)` — intended non-convergence (an unsolved reference can't compose).

PR #1296, opened 2026-06-13
feat(assembly): rotational mates — concentric + angle (kinematics Phase 1)

andymai:
## Summary

**Phase 1 of the kinematics roadmap (#1287)** — axis-based mates and the first rotational constraints, built on Phase 0's constraint composition (#1292). Addresses gaps **A1** and **A3** from #1286.

## Kernel: a face's axis of symmetry

New `getSurfaceAxis(face)` on the surface-ops interface → `{ origin, direction } | null`, surfaced as **`faceAxis()`** in `faceFns`:

| Adapter | Implementation |
|---|---|
| `occt` | Exact via `gp_Cylinder.Axis()` (location + direction) |
| `occt-wasm` (default) | Exact derivation from two surface samples — `r = \|p₁−p₂\|/\|n₁−n₂\|`, `origin = p − r·n`, `direction = n₁×n₂`. occt-wasm has no analytic `gp_Cylinder` accessor; this is exact for full **and** partial cylinders |
| `manifold` | Delegates via OCCT (matches sibling `getSurfaceCylinderData`) |
| `brepkit` | `null` — OCCT-only for now (see follow-up issue) |

## Solver: concentric + angle

- `extractEntity` now yields an **`axis`** entity from a cylindrical face or a straight edge, and a **`plane`** from a planar face (`faceAxis` doubles as the discriminator).
- Quaternion helpers + two new solvers:
  - **concentric** (axis-axis): rotate the dependent so its axis is parallel to the reference, then translate so the axes are collinear (pin-in-hole).
  - **angle** (plane-plane): rotate so the dependent's normal sits at the requested angle from the reference's.
- The reference entity is transformed into world space by its **solved pose (rotation included)**, so rotated references compose down a chain.
- Entity-type mismatches (e.g. concentric on planes) are reported `unsupported`.

## Tests

- **concentric pin-in-hole** (occt + occt-wasm; skipped on brepkit via the new `mateFns.concentricAxis` divergence) — converges with the sleeve's axis collinear with the shaft's.
- **angle relative-orientation** — the solved rotation places the dependent normal exactly 45° from the reference (verified by rotating the normal with the quaternion).
- Updated the unsupported-honesty suite: concentric/angle now **solve** for well-typed inputs; entity-type mismatches remain the unsupported cases.
- Fixed the `cylindricalFace` test helper (it selected a planar cap by Z, not the lateral face) to select by `faceAxis`.

## Verification

typecheck, eslint, boundaries, prettier, knip clean; pattern baseline restored. mateFns: occt-wasm 27 ✓ · occt 27 ✓ · brepkit 26 ✓ + 1 skipped. Public-API export list updated for `faceAxis`.

## Follow-ups
- Phase 2 — joint primitives (`jointFns.ts`)
- Phase 3 — forward kinematics + mechanism DOF

Refs #1287, #1286

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -638,6 +638,7 @@ export {
   uvCoordinates,
   normalAt,
   faceCenter,
+  faceAxis,
   classifyPointOnFace,
   outerWire,
   innerWires,
--- a/src/kernel/brepkit/geometryOps.ts
+++ b/src/kernel/brepkit/geometryOps.ts
@@ -574,6 +574,8 @@ export function makeGeometryOps(bk: BrepkitKernel) {
     createCurveAdaptor: (shape) => createCurveAdaptor(bk, shape),
     getBezierPenultimatePole: (edge) => getBezierPenultimatePole(bk, edge),
     getSurfaceCylinderData: (surface) => getSurfaceCylinderData(bk, surface),
+    // Axis-based mates (concentric) are an OCCT-only feature for now.
+    getSurfaceAxis: () => null,
     reverseSurfaceU: (surface) => reverseSurfaceU(bk, surface),
     classifyPointOnFace: (face, u, v, tolerance) => classifyPointOnFace(bk, face, u, v, tolerance),
     classifyPointRobust: (shape, point, tolerance) =>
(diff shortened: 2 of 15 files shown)


andymai:
Fixed the P1 WASM leak in 06f39c47 — `getSurfaceAxis` now wraps the `BRepAdaptor` in try-finally and deletes the intermediate `gp_Cylinder`/`gp_Ax1`/`gp_Pnt`/`gp_Dir` objects, matching the `getBezierPenultimatePole` pattern in the same file. occt concentric tests still pass.

PR #1302, opened 2026-06-13
feat(assembly): drivable joint primitives (kinematics Phase 2)

andymai:
## Summary

**Phase 2 of the kinematics roadmap (#1287)** — drivable revolute/prismatic joints built on the assembly tree, plus a shared quaternion util. Builds on Phase 0 (composition, #1292) and Phase 1 (rotational mates, #1296).

## New: `src/operations/jointFns.ts`

A **joint** connects a `parent` (reference) body to a `child` (moving) body about/along an axis, carrying one drivable parameter clamped to a range.

- `revoluteJoint(parent, child, axis, opts?)` — hinge; `value` is degrees, default range ±180.
- `prismaticJoint(parent, child, axis, opts?)` — slider; `value` is length units, default range 0..100.
- `setJointValue(joint, value)` — immutable, clamps to range.
- **`jointTransform(joint, value?)`** — the child's local rigid transform (relative to the parent) for a joint value:
  - revolute → rotation of `value`° about the axis **line**: `p ↦ R·p + (origin − R·origin)`,
  - prismatic → translation of `value` units along the axis direction.
- `addJoint(assembly, joint)` — attaches a joint to a node (new `joints?` field on `AssemblyNode`, parallel to `mates`).

`joint.value` is the stored degree of freedom, **distinct from `AssemblyNode.rotate`** (a static structural transform). A joint is conceptually sugar over the Phase-1 mates (revolute ≈ concentric + angle driver; prismatic ≈ coincident + distance driver), but the local transform is computed directly here.

## Refactor: shared quaternion util

Extracted the quaternion math (`quatRotate`, `quatFromAxisAngle`, `quatFromTo`) into **`src/utils/quaternion.ts`** (Layer 0, no internal imports) and refactored `solverAdapter` to reuse it instead of private copies. Phase 3 forward kinematics will share the same helpers. **No behavior change** — all 27 mate tests still pass.

## Tests

`tests/jointFns.test.ts` (10 tests, pure math — no kernel needed):
- construction + range clamping (initial value and `setJointValue`),
- **revolute exit criterion**: a point 1 unit off a +Z axis sweeps +X→+Y at 90°; axis-offset case (Z-axis through (5,0,0): (6,0,0)→(5,1,0)); points on the axis stay fixed; explicit-value clamping,
- prismatic translation (incl. non-unit axis normalization),
- `addJoint` immutability.

## Verification

typecheck, eslint, boundaries, prettier, knip, patterns all clean. mate+joint suites: occt-wasm 37 ✓ · occt 37 ✓ · brepkit 36 ✓ + 1 skipped. Public-API export list updated for the 5 new functions.

## Follow-up
- Phase 3 — `forwardKinematics(assembly, jointValues)` propagating joint poses down the chain + real mechanism DOF.

Refs #1287


andymai:
Both P2s addressed in 3af4a5d7:
- **Inverted range:** `makeJoint` now normalizes via `Math.min`/`Math.max`, so a `min > max` input can't push `value` below `min`. Added a regression test.
- **Prismatic origin:** documented on `prismaticJoint` that `axis.origin` is ignored (a pure translation has no anchor), unlike revolute.

Issue #1312, opened 2026-06-13
occt: getSurfaceCylinderData throws (oc.GeomAdaptor_Surface_2 is not a constructor)


andymai:
## Gap

On the **occt** kernel (opencascade.js / brepjs-opencascade WASM build), `getSurfaceCylinderData(surface)` throws:

```
TypeError: oc.GeomAdaptor_Surface_2 is not a constructor
```

The implementation (`src/kernel/occt/geometryQueryOps.ts:444`) does:

```ts
const adaptor = new oc.GeomAdaptor_Surface_2(surface);
```

but `GeomAdaptor_Surface_2` is not bound in the current `brepjs-opencascade` WASM build. `getSurfaceAxis` (`:467`) similarly uses `oc.BRepAdaptor_Surface_2` — that one works because it's bound.

## How it was masked

The only test exercising this path (`tests/curves.test.ts`, "projects curves onto a cylindrical face with original scale") wrapped the call in `try/catch` and only asserted *if a cylinder face was found* — so the throw silently made the assertion body unreachable on occt. The `manifold` kernel proxies to occt (`viaOcct`) and inherits the same break.

## Impact

`sketchOnFace` / `curvesAsEdgesOnFace(..., 'original')` (unwrapping a sketch onto a cylindrical face at true scale) fails on the occt kernel. The default kernel (occt-wasm) is now unaffected — radius/`isDirect` are derived from surface sampling (#1298).

## Done when

- occt `getSurfaceCylinderData` returns `{ radius, isDirect }` for a cylindrical surface instead of throwing (bind the right adaptor, or derive analytically).
- The `curves.cylinderUnwrapOriginal` divergence entries for `occt` and `manifold` (`tests/helpers/kernelDivergences.ts`) are removed and the assertion runs there too.

Surfaced while implementing #1298.

PR #1314, opened 2026-06-13
feat(occt-wasm): derive getSurfaceCylinderData from surface sampling

### LINKED ISSUE #1298: occt-wasm: getSurfaceCylinderData is notImplemented (no cylinder radius on default kernel)
## Gap

On the **default kernel** (occt-wasm), `getSurfaceCylinderData(surface)` throws `notImplemented` (`src/kernel/occtWasm/occtWasmAdapter.ts`), while the `occt` (opencascade.js) and `brepkit` adapters implement it. So a caller asking for a cylinder's **radius** gets a hard failure on the kernel brepjs ships by default.

```ts
getSurfaceCylinderData(_surface: KernelType): { radius: number; isDirect: boolean } | null {
  notImplemented('getSurfaceCylinderData');
}
```

## Root cause / options

The occt-wasm native module exposes no analytic `gp_Cylinder` accessor (only `surfaceType`, `surfaceNormal`, `pointOnSurface`, `getSurfaceCenterOfMass`, `surfaceCurvature`). Two ways forward:

1. **Derive in the adapter** (no upstream change): radius can be recovered the same way #1287 Phase 1 recovered the cylinder axis — two surface samples at the same height give `r = |p1-p2| / |n1-n2|`; or `r = 1 / principalCurvature` via `surfaceCurvature`. See `getSurfaceAxis` in `src/kernel/occtWasm/surfaceOps.ts` for the pattern.
2. **Upstream** the `occt-wasm` package to expose analytic surface params (cylinder radius + axis), which would also let `getSurfaceAxis` read it directly instead of deriving.

## Done when

`getSurfaceCylinderData` returns the correct `{ radius, isDirect }` on occt-wasm instead of throwing.

Surfaced while implementing #1287 Phase 1.

andymai:
## Summary

On the default kernel (**occt-wasm**), `getSurfaceCylinderData` threw `notImplemented`, so any caller asking for a cylinder's radius got a hard failure. occt-wasm exposes no analytic `gp_Cylinder` accessor, so this derives the data from the same two surface samples `getSurfaceAxis` already takes:

```
radius   = |p1 - p2| / |n1 - n2|
isDirect = n·(p - axisOrigin) > 0   // normal points radially outward
```

Both functions now share one `deriveCylinder` helper in `surfaceOps.ts`.

## Changes

- `src/kernel/occtWasm/surfaceOps.ts`: add `deriveCylinder`; `getSurfaceAxis` and the new `getSurfaceCylinderData` both read from it.
- `src/kernel/occtWasm/occtWasmAdapter.ts`: wire `getSurfaceCylinderData` to `surfaceOps` (occt-wasm uses faces as surface proxies).
- `tests/curves.test.ts`: the cylinder-unwrap test previously swallowed the `notImplemented` throw in a `try/catch` and asserted nothing. It now asserts `radius ≈ 5` and `isDirect === true`, and runs the unwrap.
- `tests/helpers/kernelDivergences.ts`: gate that assertion on `occt` and `manifold`, which hit a **pre-existing** `GeomAdaptor_Surface_2` binding bug in the brepjs-opencascade WASM build (filed as #1312).

## Verification

`tests/curves.test.ts`:
- occt-wasm (default): 20/20
- brepkit: 20/20
- occt: 19/20 (1 skipped via divergence — see #1312)
- manifold: pre-existing failures unchanged; cylinder test skipped via divergence

Closes #1298

### DIFF
--- a/src/kernel/occtWasm/occtWasmAdapter.ts
+++ b/src/kernel/occtWasm/occtWasmAdapter.ts
@@ -1508,8 +1508,9 @@ export class OcctWasmAdapter implements KernelAdapter {
     notImplemented('untrimFace');
   }
 
-  getSurfaceCylinderData(_surface: KernelType): { radius: number; isDirect: boolean } | null {
-    notImplemented('getSurfaceCylinderData');
+  getSurfaceCylinderData(surface: KernelType): { radius: number; isDirect: boolean } | null {
+    // occt-wasm uses faces as surface proxies (see extractSurfaceFromFace).
+    return surfaceOps.getSurfaceCylinderData(this.k, surface);
   }
 
   getSurfaceAxis(
--- a/src/kernel/occtWasm/surfaceOps.ts
+++ b/src/kernel/occtWasm/surfaceOps.ts
@@ -7,6 +7,7 @@
 import type { KernelShape, SurfaceType } from '@/kernel/types.js';
 import type { OcctKernelWasm, OcctWasmModule } from './occtWasmTypes.js';
 import { handle, unwrap } from './helpers.js';
+import { shapeOrientation } from './topologyOps.js';
 
 export function vertexPosition(k: OcctKernelWasm, vertex: KernelShape): [number, number, number] {
   const vec = k.vertexPosition(unwrap(vertex));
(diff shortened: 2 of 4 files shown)


andymai:
Addressed the `isDirect` semantic gap in 0db38d4d: occt-wasm's `surfaceNormal` is orientation-aware, so a REVERSED cylindrical face (inner bore) reported an inward normal and derived `isDirect=false`. `deriveCylinder` now compensates with `shapeOrientation`, so the result matches `gp_Cylinder::Direct()` regardless of face orientation — exactly the reversed-face case you flagged. Added an occt-wasm regression test on a tube's bore face (radius 2, reversed orientation → `isDirect=true`).

PR #1327, opened 2026-06-13
feat(mates): concentric axis extraction for cones, tori, and surfaces of revolution

### LINKED ISSUE #1299: Assembly mates: extend axis extraction beyond cylinders (cones, surfaces of revolution, circular edges)
## Gap

`extractEntity` (`src/operations/mateFns.ts`) currently derives an `axis` SolverEntity only from:
- a **cylindrical face** (via `faceAxis` → `getSurfaceAxis`), and
- a **straight edge** (`getCurveType === 'LINE'` → start point + tangent).

Other geometry that defines a perfectly good rotation/concentric axis is not yet supported:
- **conical faces** (`gp_Cone.Axis()`) — tapered pins/holes,
- **surfaces of revolution** and **toroidal** faces (their axis),
- **circular / elliptical edges** — a bore rim or shaft fillet edge implies an axis (normal of the circle's plane through its center).

For these, `extractEntity` returns `plane` (faces) or `null` (edges), so a `concentric` mate against them is reported unsupported.

## Suggested approach

- Extend `getSurfaceAxis` to handle cones / surfaces of revolution (occt: `gp_Cone.Axis()`, etc.; occt-wasm: same normal-sampling derivation generalizes).
- Add a `curveAxis(edge)` helper for circular/elliptical edges (center + plane normal).
- Have `extractEntity` use them so `concentric` works for the broader set.

## Done when

A `concentric` mate solves when its entities are cones, surfaces of revolution, or circular edges — not just cylinders and straight lines.

Surfaced by #1287 Phase 1.

andymai:
Completes #1299 (circular edges landed in #1325). `getSurfaceAxis` now resolves an axis for conical, toroidal, and surface-of-revolution faces, so a `concentric` mate solves against tapered pins/holes, not just cylinders and circular edges.

## Per-kernel implementation
- **occt-wasm (default):** derive the axis by sampling. The iso-v curve of a revolved surface is a circle centered on the axis, so two circle centers at different heights span it. Handles cone, torus, and general revolution. Verified exact for all three.
- **occt:** read the analytic axis where bound (cylinder, surface of revolution). `gp_Cone`/`gp_Torus` are not bound in the shipped opencascade.js WASM build, so those accessors throw; that is caught and returns null rather than failing. Refactored to a small `surfaceAxisAx1` helper.
- **brepkit:** read cone apex + axis from `getAnalyticSurfaceParams`. Torus params carry only a center (no axis vector), so a toroidal axis is not recoverable there.
- **manifold:** mesh kernel, so cones tessellate to planar facets with no analytic cone face (N/A).

## Verification
- New concentric mate test on two conical faces: passes on occt-wasm and brepkit; skipped on occt (`gp_Cone` unbound) and manifold (no analytic face) via divergences.
- New `curveAxis` arc test (addresses Greptile's note on #1325): a 90-degree arc resolves the same center as the full circle. Skipped on brepkit, which types arcs as NURBS rather than CIRCLE.
- `mateFns.test.ts` + `curveFns.test.ts`: occt-wasm 44/44, brepkit 43/44 (1 skip), occt 43/44 (1 skip).

Closes #1299.

### DIFF
--- a/src/kernel/brepkit/geometryOps.ts
+++ b/src/kernel/brepkit/geometryOps.ts
@@ -399,9 +399,16 @@ export function getSurfaceAxis(
   face: KernelShape
 ): { origin: [number, number, number]; direction: [number, number, number] } | null {
   const params = JSON.parse(bk.getAnalyticSurfaceParams(unwrap(face, 'face')));
-  if (params.type !== 'cylinder') return null;
-  const [ox, oy, oz] = params.origin;
-  const [ax, ay, az] = params.axis;
+  // Cylinder params expose origin+axis; cone params expose apex+axis (the apex
+  // is a point on the axis). Torus params carry only a center, no axis vector,
+  // so a toroidal axis is not recoverable here.
+  // params is `any` from JSON.parse; cylinder exposes origin, cone exposes apex.
+
+  const point =
+    params.type === 'cylinder' ? params.origin : params.type === 'cone' ? params.apex : null;
+  if (!point || !params.axis) return null;
+  const [ox, oy, oz] = point as [number, number, number];
+  const [ax, ay, az] = params.axis as [number, number, number];
   const len = Math.hypot(ax, ay, az);
   if (len < 1e-12) return null;
   return {
--- a/src/kernel/occt/geometryQueryOps.ts
+++ b/src/kernel/occt/geometryQueryOps.ts
@@ -463,7 +463,34 @@ export function getSurfaceCylinderData(
   return result;
 }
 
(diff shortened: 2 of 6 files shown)


andymai:
Addressed both review points in eb332243:
1. occt `getSurfaceAxis` now wraps `axis`/`loc`/`dir` in try-finally so they're released if `Location()`/`Direction()` throws.
2. occt-wasm `deriveAxisBySampling` samples three v-levels and takes the farthest-apart center pair, so a v-symmetric partial torus (half-torus) no longer collapses to coincident centers.

PR #1377, opened 2026-06-15
perf(brepkit): use binary grouped tessellation (no JSON marshaling)

andymai:
## Summary

The brepkit mesh adapter round-tripped every mesh through a **JSON string**. `tessellateSolidGrouped` returned JSON; we `JSON.parse`d it and re-packed the number arrays into typed arrays (`new Float32Array(numberArray)`). For a sphere at tol 0.01 (56k triangles → a **4.2 MB JSON string**) that added **~46 ms** on top of the ~35 ms of actual tessellation — roughly doubling meshing cost.

This prefers the kernel's binary `tessellateSolidGroupedBinary` when present: it returns packed `Float32Array`/`Uint32Array` buffers directly, so the mesh crosses the WASM boundary as bulk memory copies — no JSON. Falls back to the JSON path (`meshSolidGroupedJson`) on kernels without the binary API.

## Measured (kernel-comparison suite)

| op | before | after |
|---|---|---|
| mesh sphere (56k tris) | 81 ms | **33 ms** |
| cut(box,cyl) ×10 | 106 ms | **62 ms** |
| multi-boolean (16 ops) | 13 ms | **7 ms** |

## Compatibility

- The binary method is **feature-detected** (`typeof bk.tessellateSolidGroupedBinary === 'function'`). Older kernels keep working via the JSON fallback, so this is safe to merge ahead of the brepkit-wasm release.
- Activates the fast path once the brepkit-wasm dependency is bumped to a release containing `tessellateSolidGroupedBinary` (brepkit PR andymai/brepkit#817).

Full pre-commit suite green locally: typecheck, layer boundaries, eslint/prettier, **3701 tests pass**.

### DIFF
--- a/scripts/sync-brepkit-types.ts
+++ b/scripts/sync-brepkit-types.ts
@@ -37,9 +37,12 @@ function snakeToCamel(s: string): string {
 function convertParams(raw: string): string {
   let result = raw;
   // Convert param names from snake_case to camelCase
-  result = result.replace(/\b([a-z][a-z0-9_]*)\s*([\?]?\s*:)/g, (_, name: string, colon: string) => {
-    return snakeToCamel(name) + colon;
-  });
+  result = result.replace(
+    /\b([a-z][a-z0-9_]*)\s*([\?]?\s*:)/g,
+    (_, name: string, colon: string) => {
+      return snakeToCamel(name) + colon;
+    }
+  );
   // Widen typed array params to also accept plain number[]:
   // Float64Array → Float64Array | number[]
   // Uint32Array → Uint32Array | number[]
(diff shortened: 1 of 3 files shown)


andymai:
Thanks for the review — addressed in 42bc1b31:

- ✅ **Type-sync registration** — `tessellateSolidGroupedBinary` added to the Tessellation list in `sync-brepkit-types.ts`; both forward-declarations in `brepkitWasmTypes.ts` are now documented (why they're hand-added and optional).

**On returning un-copied typed arrays** — these are owned copies, not views. For a `Vec<f32>`/`Vec<u32>` return, wasm-bindgen generates `getArray…FromWasm0(ptr, len).slice()` and then `__wbindgen_free`s the Rust buffer, so the returned typed array owns its data and is safe to hold past the call. Wrapping it in `new Float32Array(...)` would just reintroduce the copy this change is removing — the defensive copies in `meshSingleFace`/`meshEdgesAll` are themselves copying already-copied data. So skipping the re-copy here is correct, not a convention break.

**On test coverage** — the face-group assembly loop is identical to the JSON path the suite exercises today; the binary-specific delta is the f64→f32 conversion, now unit-tested on the Rust side (andymai/brepkit#817). End-to-end, the binary path was validated against a local kernel build via the kernel-comparison bench. A committed integration test follows once the brepkit-wasm dep is bumped and the binary path becomes the default (the `typeof` guard currently routes all CI runs through the JSON fallback).

PR #1380, opened 2026-06-15
fix(verify): reap orphaned sandbox processes on timeout and host shutdown

andymai:
## Problem

Sandboxed `run_program` / `export_part` verify runs could leak `node` processes that spin a CPU core indefinitely. Observed in the wild: dozens of orphaned `brepjs-verify … --check` processes up to 13h old, saturating every core.

## Root cause

The sandbox's only runaway protection is the per-run wall-clock timeout (`runProgram.ts`), enforced via `child_process`' `timeout` + `killSignal`. Two ways it failed to actually kill the work:

- **Mode A — timeout signals the wrong process.** In dev/test the CLI runs as `npx tsx <main.ts>`, so the process that executes the part (and spins on a CPU-bound OCCT op) is a *grandchild* behind npx+tsx. `child_process` signals only the **direct child** (`npx`), so a fired timeout SIGKILLed npx and orphaned the still-spinning grandchild, which reparented to init and kept burning a core.
- **Mode B — the timer dies with the host.** The timeout lives in the host process (the MCP server). If the server is stopped — the agent disconnecting — before a run's budget elapses, the timer never fires and the whole `detached` sandbox group is orphaned. This is what produced the 13h-old runaways.

## Fix

- **Process-group isolation + group kill.** Spawn the CLI `detached` (its own process group) and SIGKILL the **whole group** (`process.kill(-pid)`) on timeout/output-cap, so the entire `npx → tsx → node` tree is reaped. `CliOutcome` semantics (stdout/stderr/timedOut/outputTooLarge/exitCode) are unchanged.
- **Host-shutdown reaping.** Track each run's group in a registry; install `exit`/`SIGTERM`/`SIGINT` hooks (`installSandboxShutdownHandlers`, wired into the MCP server entrypoint) that SIGKILL every in-flight group on the way down. Exposes `killActiveSandboxes()` for explicit reaping.

### Known residual
A hard `SIGKILL` of the host can't be trapped (no handler runs); that case needs the kernel's `PR_SET_PDEATHSIG`, which Node doesn't expose without a native addon. Documented in the `installSandboxShutdownHandlers` docstring.

## Tests (TDD — each watched fail first, for the right reason)

- Mode A: records the **leaf** PID inside the runaway, asserts it's dead after the timeout (the existing timeout test only checked the return value, which `execFile` reported correctly even while leaking).
- Mode B: spawns a host fixture (direct `node`, mirroring the built server), lets it start a runaway, SIGTERMs it, asserts the sandbox leaf is reaped rather than orphaned.

Full `runProgram` suite + adjacent sandbox/entrypoint/cli tests green; typecheck + lint clean; full `test:ci` (4116 tests) green on push.

### DIFF
--- a/packages/brepjs-verify/src/mcp/server.ts
+++ b/packages/brepjs-verify/src/mcp/server.ts
@@ -24,6 +24,11 @@ import {
   runProgramTool,
   type ExportPartToolArgs,
 } from './tools.js';
+import { installSandboxShutdownHandlers } from '../sandbox/runProgram.js';
+
+// Reap any in-flight sandbox process groups if this server is stopped mid-build (the agent
+// disconnecting), so a runaway part can't outlive the server and burn a core indefinitely.
+installSandboxShutdownHandlers();
 
 // Read the version from package.json at runtime so it tracks the package (this module sits two
 // levels below the package root in both src/ and dist/). NB: don't use
--- a/packages/brepjs-verify/src/sandbox/runProgram.ts
+++ b/packages/brepjs-verify/src/sandbox/runProgram.ts
@@ -8,16 +8,13 @@
  * live WASM handles), and the temp program directory is always cleaned up.
  */
 
-import { execFile } from 'node:child_process';
-import { promisify } from 'node:util';
+import { spawn } from 'node:child_process';
 import { mkdtemp, writeFile, rm } from 'node:fs/promises';
 import { existsSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import type { VerifyReport } from '../verify/report.js';
-
-const execFileAsync = promisify(execFile);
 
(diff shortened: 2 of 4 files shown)


andymai:
Addressed the three P2 findings from the Greptile review in `63361c0`:

1. **stdout cap early-out** (`runProgram.ts`) — added `if (outputTooLarge) return;` at the top of the `data` handler so a condemned run no longer re-fires `SIGKILL` on every subsequent buffered chunk.
2. **stderr byte accounting** (`runProgram.ts`) — cap now tracks `chunk.length` (bytes) instead of `String.length` (UTF-16 code units), matching stdout so non-ASCII output can't slip past the 8 MB budget.
3. **Mode A test determinism** (`runProgram.test.ts`) — replaced the fixed 1 s wait with the same poll-until-reaped loop Mode B uses, avoiding the dead-but-unreaped flake window on loaded CI.

Verified locally: lint + typecheck clean, full `runProgram` suite (6/6) green; pre-push `test:ci` (4116 tests) green.

PR #1414, opened 2026-06-15
fix(deps): point brepjs-verify's brepjs-viewer devDep at the workspace

andymai:
## Summary

**Unblocks `main` CI.** Every job has been failing at `npm ci` with:
```
npm error notarget No matching version found for brepjs-viewer@0.2.1
```

`packages/brepjs-verify` pinned `brepjs-viewer@0.2.1` (devDependency), but the workspace `brepjs-viewer` is `0.2.0` and only `0.1.0` is published — so npm couldn't satisfy the exact `0.2.1` from the workspace and fell back to the registry, which 404s. A release-please version desync from the recent auto-publish-viewer work.

## Fix

Point the **devDependency** at the workspace (`"*"`, exactly as `apps/playground` already does), so it resolves to the local `packages/brepjs-viewer` instead of the registry. It's dev-only, so the loose range never reaches published consumers of `brepjs-verify`.

## Verification

- `npm ci --dry-run` resolves cleanly (was ETARGET).
- Lockfile regenerated; `brepjs-viewer@0.2.1` reference removed; `@emnapi` entries preserved (npm 11.15).
- Pre-commit green (typecheck + occt-wasm changed suite: 4188 passed).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5614,8 +5614,7 @@
     },
     "node_modules/brepjs": {
       "resolved": "",
-      "link": true,
-      "version": "18.83.1"
+      "link": true
     },
     "node_modules/brepjs-bim": {
       "resolved": "packages/brepjs-bim",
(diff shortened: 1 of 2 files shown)


andymai:
Closing as duplicate of #1408, which already applies this exact fix (unpin brepjs-verify's brepjs-viewer devDep to the workspace). Both are blocked by the same ci-pass/size deadlock — the size-limit-action can't build the broken base main for its delta comparison.

PR #1439, opened 2026-06-16
docs(landing): rebuild brepjs.dev landing — code-as-CAD hero + copy refresh

andymai:
## What

Rebuilds the brepjs.dev landing page as a bespoke page rendered inside VitePress (a `landing: true` frontmatter intercept in `theme/Layout.vue` → `Landing.vue`; all other routes keep DefaultTheme). Design tokens are scoped to `.landing` so nothing bleeds into the docs theme.

## Highlights

- **Code-as-CAD hero** (`CodeCadHero.vue` + `codeCadRenderer.ts`): types out real TypeScript character-by-character and builds a **1×1 Gridfinity bin** — the heritage part — step by step (socket foot → hollow body → stacking lip). Geometry is pre-baked by running occt-wasm in Node (`scripts/genHeroFrames.ts`) and rendered with Three.js as face meshes **plus exact B-Rep edges**. Orthographic iso view + manual orbit/zoom, play-once with a Replay button, and simulated TypeScript IntelliSense (completion dwell + hover signatures), including a `drawRoundedRectangle` completion on the first brepjs call.
- **Real syntax highlighting** in every editor via a shared tokenizer (`codeHighlight.ts`) — the same source drives the hero's typing and the static cards.
- **Honest section visuals**: exact-vs-faceted and the verify section render real kernel geometry (`scripts/genLandingImages.ts`); the type-safety card shows a real `tsc` error.
- **Copy refresh**: broadened the B-Rep lens (mechanical CAD geometry, parts → assemblies) across the landing, `README.md`, and `introduction/why-brepjs.md`; accurate footer tagline (brepjs is a library, not a kernel); the closing "Authoring with AI?" line is now an actionable link to the verify-skill guide.
- **Accessibility**: AA contrast, adequate target sizes, landmarks/headings, no horizontal overflow; reduced-motion fallback for the hero.
- **Playground**: adds a parametric Gridfinity bin example (`cols` × `rows` × `heightUnits`, one socket foot per cell) as the flagship mechanical example, with a generated thumbnail.

## Verification

- `npm run build --workspace=apps/docs` is clean; landing verified at desktop + mobile widths via headless capture.
- Playground typechecks (`tsc -b`); the new example builds, meshes, and renders on the real kernel — 1×1 (default) and 3×2 both confirmed.


andymai:
Thanks @greptile — triaged all three:

**1. `codeCadRenderer.ts` — `shown` not reset on replay (P1): Fixed** in de910ed9. Confirmed real: `restart()` → `hide()` zeroed the slot targets but left the `shown` guard, so clicking Replay during a step's dwell made the next `showStep(0)` early-return and the socket-foot phase never re-faded-in. `hide()` now also sets `shown = -1`. Verified end-to-end headlessly — reproduced the race (Replay during the step-0 dwell) and confirmed the socket foot re-renders on the replay.

**2. `mechanical.ts` — WASM disposal in the cell loop (P1): Working as intended.** This matches the deliberate house style for playground examples — none of the existing examples manually dispose loop temporaries (e.g. the axial fan's `bladeRing.push(rotate(makeBlade(), …))`, the GT2 pulley, the fluted knob all do the same). Examples optimize for teaching the shape API; per-run kernel memory is handled by the worker + the core `FinalizationRegistry`, and the `disposal-required` pattern checker runs over `src/` (green), not examples. Adding `using`/`DisposalScope` to only this one would make the flagship inconsistent with the rest of the corpus, so leaving as-is.

**3. `genHeroFrames.ts` / `genLandingImages.ts` — disposal in build scripts (P2): Won't fix.** As you note, the practical impact is nil — these are one-shot scripts that exit immediately, internal build tooling (not shipped, not user-facing shape examples), and both follow the same pattern. Not worth the churn.

PR #1443, opened 2026-06-16
feat: unified sketch chainable surface + playground example type-check guard

andymai:
## What & why

Started from two playground-landing asks and grew to fix the root causes plus add a regression guard:

1. **Hero "Open in Playground"** now opens the broader **parametric gridfinity example** (cols × rows × height) instead of the inline 1×1 bin. The typed animation is unchanged.
2. **Fixed the red squiggles** in the playground examples — and built a guard so they can't come back.

### Sketch chainable surface (the `loftWith` squiggle)
`Drawing.sketchOnPlane()` returns `SketchInterface | Sketches`, but the union was incomplete:
- `Sketches` lacked `loftWith`/`sweepSketch`/`face`; `SketchInterface` lacked `sweepSketch`.
- `loftWith`/`sweepSketch` used `this`-polymorphic params, so union values weren't accepted as arguments either.

Fix: `Sketches implements SketchInterface`, add `sweepSketch` to the interface, widen loft/sweep targets from `this` → `SketchInterface`. Multi-profile loft/sweep on `Sketches` throws a clear error (it's genuinely ambiguous). Non-breaking (additive + widened params).

### Batch booleans (the `fuseAll`/`ValidSolid` squiggle)
Pairwise `fuse`/`cut` accept any `Shape3D`, but batch `fuseAll`/`cutAll` required `ValidSolid[]` (or `{ unsafe: true }`) — so fusing extruded/lofted/cut results squiggled. Added `Shapeable<Shape3D>` wrappers in `api.ts` (matching `fuse`/`cut`), restoring consistency.

### convexHull
Annotated the axial-fan `section` helper to return `Vec3` tuples.

## New: example type-check guard
`apps/playground/scripts/checkExamples.ts` type-checks every example's `code` against the **exact** ambient declarations the Monaco editor injects, with the editor's compiler options. Wired into the playground `build` (so CI's `playground-build` job + Vercel enforce it). The ambient→module transform is now shared between the editor and the guard so they can't drift.

This guard immediately caught a latent **ambient-types generator bug**: the experimental implicit/SDF domain re-exports primitive names (`cylinder`/`box`/…) aliased as `sdfCylinder`, but the generator emitted them under their local name, overwriting the core solid primitives. Excluded the implicit domain from the playground ambient types (no example/doc uses it) and regenerated — which also resyncs the long-stale ambient `.d.ts` with the current API.

## Verification
- `npm run validate` green (typecheck, lint, boundaries, format, 4210 changed-file tests).
- New `tests/sketchChainable.test.ts` (loft / fuseAll-of-Shape3D / sweep) passes across all 4 kernels.
- `npm run check:examples` → all 15 examples type-check cleanly.

## Notes for review
- The large `brepjs-ambient.d.ts` diff is the regeneration (it was stale since before the implicit domain landed).

### DIFF
--- a/apps/docs/.vitepress/theme/components/CodeCadHero.vue
+++ b/apps/docs/.vitepress/theme/components/CodeCadHero.vue
@@ -5,8 +5,9 @@ import { encodeCode } from '../playgroundLink';
 import { tokenize, tokensToHtml } from './codeHighlight';
 import type { CodeCadHandle, HeroFramesData } from './codeCadRenderer';
 
-// The program the panel types out, and what "Open in Playground" carries — a
-// real, runnable 1×1 Gridfinity bin. Mirrors scripts/genHeroFrames.ts.
+// The program the panel types out — a real, runnable 1×1 Gridfinity bin.
+// Mirrors scripts/genHeroFrames.ts. "Open in Playground" carries the broader
+// parametric example below, not this animation snippet.
 const PROGRAM = `import { drawRoundedRectangle, cut, fuse, unwrap } from 'brepjs/quick';
 
 const [W, WALL, H] = [42 - 0.5, 1.2, 3 * 7]; // 1×1 bin, 3 units tall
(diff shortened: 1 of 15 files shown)


andymai:
Addressed Greptile's review in `8244c2d`:

- **P1 — `CompoundSketch.loftWith` not widened:** widened the param to `SketchInterface | SketchInterface[]` to match the interface, with a runtime guard that the target is a `CompoundSketch` (else a clear `bug()`). Closes the soundness gap where a `CompoundSketch` held via a `SketchInterface` reference could be lofted with a plain `Sketch` and hit a `TypeError`.
- **P2 — `asSketch` cast:** typed the piece array as `Array<Sketch | CompoundSketch>` instead of casting to `Sketch[]` (matches `Sketches.sketches`); `first instanceof Sketch` still narrows the return.
- **P2 — hero `PLAYGROUND_PROGRAM` not guarded:** extended `checkExamples.ts` to also type-check the docs landing hero's hand-maintained program (now "16 snippets: examples + docs hero"). It errors loudly if the constant is renamed/moved, so the cross-app mirror can't silently drift.

PR #1464, opened 2026-06-17
feat(playground): make truss, I-beam, and enclosure examples true to life

andymai:
Reworks three mechanical/BIM playground examples that read as obviously synthetic, plus the small `brepjs-bim` feature one of them needs.

### `feat(bim)`: optional root fillets on I-beam profiles
Hot-rolled wide-flange sections always carry a **root fillet** where the web meets the flanges (and `IfcIShapeProfileDef` stores it as `FilletRadius`). Added an optional `filletRadius` to the `I_BEAM` profile:
- tessellate the four root corners into arcs in `profileToPolygon`
- account for the added material in `profileCrossSectionArea` (`4·r²(1−π/4)`)
- emit it to IFC (the `FilletRadius` slot was previously hard-`null`)
- tests in `profile.test.ts` + `beamFns.test.ts`

### `feat(playground)`: three examples made true-to-life
- **Cube-truss frame** — was a blocky stack of box-section cubes with flat slab braces. Now a square-section **tubular space frame**: round chord tubes, a Warren zig-zag of round web braces, and spherical gusset nodes (returned as an array of solids, like the curtain-wall example — no fragile N-way fusing).
- **Steel I-beam** — now passes `filletRadius` so the web/flange junctions show the rolled root fillets; ambient types regenerated.
- **Project enclosure** — fixed a latent bug where the lid cavity removed the entire cap (a capless frame); it's now a proper hollow shell. Added tapped corner bosses (modelled thread) + aligned lid clearance holes so the halves actually screw together. Tapping a standalone boss then fusing copies brought the build from ~17s down to ~3s.
- **RF enclosure** — added lid clearance holes aligned to the standoff bores so the self-tapping screws mate the two halves.

## Verification
- `brepjs-bim` suite: 575 passing; package typecheck clean
- `check:examples`: all 40 snippets type-check
- A temporary kernel-backed smoke test confirmed all three reworked builders produce valid solids (since removed)

## Notes
- Comments in the example `code` strings were kept terse per feedback (read in a small Monaco pane).
- The modelled enclosure threads sit inside ~1.5 mm bores (near-invisible at enclosure scale) but were kept since they were explicitly requested; build time is back to ~3s.

### DIFF
--- a/apps/playground/src/lib/examples/bim.ts
+++ b/apps/playground/src/lib/examples/bim.ts
@@ -11,13 +11,12 @@ export const BIM_EXAMPLES: readonly Example[] = [
     id: 'bim-steel-beam',
     label: 'Steel I-Beam',
     description:
-      'A parametric structural-steel wide-flange (I-beam) authored through a BimModel. The element carries a brepjs solid for display and the model serializes to IFC.',
+      'A parametric structural-steel wide-flange (I-beam) authored through a BimModel, complete with the rolled root fillets where the web meets the flanges. The element carries a brepjs solid for display and the model serializes to IFC.',
     code: `import { BimModel } from 'brepjs-bim';
 
-// A parametric structural-steel I-beam (wide-flange section), authored through
-// the BIM model rather than as raw geometry. The element carries a brepjs solid
-// (shown here); the same model serializes to a real IFC file via toIfc(model).
-// Tune the length and the I_BEAM section dimensions.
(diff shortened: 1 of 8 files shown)


andymai:
Addressed the P2 in c4800319: `filletArc` now falls back to the sharp vertex when the corner angle is outside a small band (near 0 or near π), so the `tan(α/2)` / bisector NaN paths can't propagate into the kernel from a future caller. The 90° I-beam roots are unaffected (tests still pass).

PR #1515, opened 2026-06-20
fix(brepjs-verify): heal sketch-extrude→modifier ValidSolid type gap

andymai:
## What

Heals a recurring **first-try type failure** the `brepjs-verify` skill surfaced under a clean-room author sweep: a shape born from a **2D-sketch `.extrude()`/`.revolve()`** is typed `Shape3D`, but `fillet`/`shell`/`offset` require `Shapeable<ValidSolid>`, and `validSolid()` itself takes a concrete `Solid` — so the union won't go straight in. Two independent clean-room authors hit `TS2345` on this exact path.

The skill mis-attributed the cause to **booleans** ("after a boolean, validate first") and prescribed a bare `validSolid(...)` wrap that is *itself* a `TS2345`. Ground truth (shipped `dist/**/*.d.ts`):

- `box`/`cylinder`/`cone` → `ValidSolid`; `cut`/`fuse` are `<T extends Shape3D>(base: Shapeable<T>): Result<T>` and **preserve** validity — a boolean rooted at a `ValidSolid` stays valid.
- `Sketch.extrude()` → `Shape3D` (the union, incl. `Shell`). That's the real trigger.
- `fillet`/`shell` → `Shapeable<ValidSolid>`; `validSolid(solid: Solid)` → needs a concrete `Solid`. So the lift needs an `isSolid` narrow first.

### Changes (prose only)
- **SKILL.md** Hard rules: name the `TS2345` code, the sketch-extrude origin, and the two-step `isSolid` → `validSolid` lift (or just build from a primitive).
- **modifiers.md**: correct the misleading header + pitfall; add the lift snippet.

## Verification — RED→GREEN
- **Deterministic micro-test:** bare `validSolid(<Shape3D>)` fails `TS2345: 'Shape3D' not assignable to 'Solid'. Type 'Shell' is not assignable…` → `isSolid` + `validSolid` passes `ok:true`, `shapeType: Solid`.
- **Causal re-author:** a fresh clean-room author (reading the patched skill) reproduced `wall-mount-flanged-junction-box` and **quoted the new rule verbatim** — "narrow with `isSolid`, then lift" — wrapped it in a `roundedPrism()` helper, and cleared the exact `TS2345` that failed it before. It also discarded the banned `as unknown as Solid` double-cast in favor of the prescribed lift.

## Two-signal scorecard

```
model=opus-4.8  brepjs=18.97.0  skill=brepjs-verify@0.28.0  units=mm
scope=hardest (23 of 41 playground parts; basics skipped) — first-try% is a LOWER BOUND

VALIDITY (auto signal, full sweep of the 23 hardest)
  first-try valid : 17/23  (73.9%)
  eventual  valid : 23/23  (100%)
  lift            : +26.1 pp (6 parts recovered within ≤4 attempts)

FAILURE-MODE BREAKDOWN (6 first-try fails)
  TS2345  sketch-extrude→modifier ValidSolid lift   2   tripod-rc2-plate, wall-mount   ← HEALED
  TS2339  polygon().extrude() (use drawPolysides)   1   vented-louvre-case             ← deferred (1×)
  TS2740  Drawing vs DrawingPen annotation          1   d-sub-connector                ← deferred (self-inflicted)
  ASSERT  own bounds (proud/rotation extent)        2   flanged-tee, conical-funnel    ← not a skill gap (step-4 covers)

DESIGN (blind judge) — SAMPLED, not full corpus
  tripod-rc2-plate : designed   (chamfered prism, both end notches, counterbored hole legible)
  wall-mount       : partial    (enclosure/ridge/lid clear; one ear reads round-holed — first-valid quality, polish pass skipped)
  all other parts  : judge:—    (coverage gap, see caveats)
```

### Judge caveats (no silent caps)
- The design signal is **a 2-part sample** (the two healed parts), not the full corpus — every other part is `judge:—`.
- It is an **absolute** read (part vs description), **not** the pairwise author-vs-reference protocol: the playground reference code fails to render under the verify CLI (its `brepjs/quick` chaining surface doesn't survive a naive `quick→brepjs` swap), so the A/B pairwise judge could not run this pass.
- Clean-room authors stop at `ok:true` with no `--snapshot`, so this measures **first-valid** design quality, not the skill's polished ceiling.

## Deferred (single-incident — listed, not fixed, per the recurrence rule)
- **`polygon().extrude()` → `TS2339`** (vented-louvre): `polygon()` returns `OrientedFace` (no `.extrude()`); the extrudable regular-polygon path is `drawPolysides(r, sides).sketchOnPlane().extrude()`, which is absent from `sketching-2d.md`. Real omission, 1 incident.
- **`DrawingPen.fillet`/`.chamfer` → `TS2339`** (surfaced in the tripod re-author): `llms-full.txt:207` advertises bare `fillet(radius)`/`chamfer(radius)` as pen methods, but the typed `DrawingPen` surface only exposes `closeWithCustomCorner(radius, 'fillet'|'chamfer')`. Real doc-vs-type contradiction in a bundled reference, but 1 incident — let it earn a fix by recurring.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (`/heal-skill`)


andymai:
Re Greptile's 4/5 caveat ("couldn't directly inspect `dist/*.d.ts`") — here is the ground truth the diagnosis was verified against, from the shipped declarations:

```
dist/sketching/drawing.d.ts:90  sketchOnPlane(inputPlane: Plane): SketchInterface | Sketches
dist/sketching/sketch.d.ts:33   extrude(distance, config?): Shape3D          // the union — incl. Shell
dist/topology/api.d.ts:83       fillet<T extends ValidSolid>(shape: Shapeable<T>, ...): Result<T>
dist/topology/api.d.ts:91       shell<T extends ValidSolid>(shape: Shapeable<T>, ...): Result<T>
dist/core/validityTypes.d.ts:98 validSolid(solid: Solid): Result<ValidSolid, string>   // concrete Solid, not Shape3D
dist/topology/primitiveFns.d.ts box/cylinder/cone(...): ValidSolid                       // primitives are valid
dist/topology/api.d.ts:58       cut<T extends Shape3D>(base: Shapeable<T>, ...): Result<T>  // preserves T
dist/core/shapeTypes.d.ts:101   isSolid(s): s is Solid
```

And the empirical RED→GREEN that backs the prose (run via the verify CLI `--check`):

```
RED  — bare validSolid(<Shape3D>):  ok:false
  TS2345: Argument of type 'Shape3D' is not assignable to parameter of type 'Solid'.
    Type 'Shell' is not assignable to type 'Solid'.
GREEN — isSolid(x) ? unwrap(validSolid(x)) : …:  ok:true, shapeType: Solid
```

So every type claim in the diff is grounded in the shipped `.d.ts` and reproduced empirically, not inferred.

PR #1524, opened 2026-06-20
fix(brepjs-verify): correct spur-gear example flanks + add root fillet

andymai:
## Summary

Fixes the `brepjs-verify` skill's `spur-gear.brep.ts` example — it built a valid solid but with **inverted flanks**, so its teeth widened toward the tip (a dovetail) and could not mesh.

## The bug

The `+angle` flank used the un-mirrored involute with `offset = halfTooth − phiPitch`, which makes the flank *diverge* toward the tip → teeth wider at the tip than the base. A dovetail tooth can't roll through a mating space, so two such gears jam (17–68 mm³ interference even at the best phase). A single gear still passes `isValidSolid` and has the correct OD, so it slipped through — only a meshing test catches it.

## The fix

1. **Corrected flank** — the `+angle` flank now uses the *mirrored* involute with `offset = halfTooth + phiPitch`, so each tooth narrows to its tip like a real cut gear.
2. **Root fillet** — replaced the sharp radial root with a circular fillet (tangent to the flank at the base circle, run to the shared space-centre point), which clears the mating tooth tip below the base circle and reads as a real cut root.

Also updates `spur-gear.expected.json` (volume/area shift with the new profile) and documents both points in `references/gears.md`.

## Verification

- The corrected single gear builds valid (`ok:true`, OD 44 mm, filleted roots).
- **Two of the corrected gears mesh at 0.0 mm³** at the correct phase (swept with `intersect`+`measureVolume`), with a clean backlash window — vs 17–68 mm³ before.

Found while building the planetary-gear-reducer playground example (#1519), which inherited this same bug.

### DIFF
--- a/packages/brepjs-verify/skill/examples/spur-gear.brep.ts
+++ b/packages/brepjs-verify/skill/examples/spur-gear.brep.ts
@@ -1,20 +1,25 @@
-// gear.brep.ts — involute spur gear (math ported from BOSL2 gears.scad for parity)
-// One closed polygon of all teeth -> extrude. Avoids per-tooth booleans (robust on occt-wasm).
+// spur-gear.brep.ts — involute spur gear that MESHES (math ported from BOSL2 gears.scad).
+// One closed polygon of all teeth -> extrude (robust on occt-wasm; no per-tooth booleans).
+// Two correctness points a "looks like a gear" build gets wrong:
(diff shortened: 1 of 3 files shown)


andymai:
Superseded by #1528, which re-homes this exact fix onto the `brepjs-cad` skill path. This PR targeted `packages/brepjs-verify/skill/…`, deleted by the brepjs-verify→brepjs-cad rename (#1520), so it could not merge. Thanks!

PR #1531, opened 2026-06-20
feat(brepjs-cad): deterministic body/interference metrics for the design judge (Phase 1)

andymai:
## What

Phase 1 of the design-judge upgrade ([reviewed spec](https://github.com/andymai/brepjs/pull/1530)): give the judge **deterministic kernel facts** it can't see in a render, so it stops inferring a part's structure from pixels. No renderer changes — this is the cheapest, highest-coverage layer.

## Why

A survey of all 51 playground examples: the mechanical corpus is **91% multi-body, 88% internal-feature**, and **32/51 parts are both** — exactly what an exterior silhouette is blind to (the current judge graded cleanly on only 1 of 14 hardest parts). Determinism is objective, cheap, and works where vision can't.

## What's in it

- **Report fields** (`src/verify/report.ts`): optional `bodies`, `bodyRelations`, `manufacturability` on `VerifyReport` (additive — existing consumers unaffected).
- **Metrics** (`src/verify/manufacturability.ts`): per-body volume/bounds/validity + a pairwise **interference matrix** (`separate` vs touching/overlapping `interfering`). Reuses the kernel's `checkInterference` (which already does AABB pre-rejection) but drives the pair loop itself for a **wall-clock budget** — `BRepExtrema` distance is pathologically slow on high-face gears, so the matrix caps (`MAX_BODIES`, 15s budget) and flags `relationsTruncated` rather than hanging. Verified on the 18-body planetary: budget held, flagged truncation, no hang.
- **Opt-in** (`--metrics`): metrics compute only behind `brep verify --metrics` (the `snapshot` subcommand sets it). Plain `--check` — the clean-room author hot path — is byte-identical to before.
- **Judge fusion**: `bench/metrics.ts` digests the report; the SDK judge (`judge.ts`) takes it via `JudgeInput.metrics` + a new `manufacturable`/`usedMetrics` verdict axis with a guidance paragraph (reconcile body count, decide intended-assembly vs collision); the **$0 blind protocol** (`blind-judge.md`) pastes a **symmetric, label-free** per-render facts line (doesn't de-blind).
- **Scorecard**: `SCHEMA_VERSION`→2, per-result `mfg:⚠` + a `manufacturable %` tally; `loop.ts`/`live.ts` thread the digest + axis.

## Deliberately deferred (honest scope)

- `minRadius` / `internalCylinderCount` — a curvature-sign probe showed the kernel's sign does **not** separate a concave bore from a convex boss (both came back negative on a tube), and fillet faces pollute `minRadius`. Shipping a flaky metric would mislead the heal loop, so deferred until a proper internal-feature test lands.
- `touching`/`nested` sub-classification + overlap-volume (needs a boolean-volume tier), per-body color/exploded render, Set-of-Marks, aimed sections — later phases per the spec.

## Verification

- `tests/manufacturability.test.ts`: interfering-vs-separate classification, coplanar-touching → interfering, **metrics absent without `--metrics`**, single-solid case, pure `digestMetrics`/`formatScorecard`.
- Full package suite **164/164**; typecheck/eslint/prettier/knip/boundaries all clean.
- E2E: `brep verify --metrics --json` populates the fields on a 3-body compound (A&B interfering, C separate) and on the planetary (budget-capped, truncation flagged).

### DIFF
--- a/packages/brepjs-cad/bench/blind-judge.md
+++ b/packages/brepjs-cad/bench/blind-judge.md
@@ -15,12 +15,15 @@ is _pairwise_ (author vs reference) and _blind_ (labels stripped), and costs not
 One judge subagent (general-purpose — it must `Read` the PNGs) per part. Its **entire input**: the
 part's NL `description` + two rendered images each — **iso + one detail view, the SAME view pair for
 both renders** so the A/B comparison is fair (pick the detail view — front/top/right — that best
-exposes the features the description names) — labeled only **A** and **B** + the rubric below. It is told NOTHING about which render is the skill's clean-room output vs
-the playground reference, and never reads any `.brep.ts`, the heal, the orchestration, or
+exposes the features the description names) — labeled only **A** and **B**, **a one-line "measured
+facts" string per render** (body count + interference relations — the ground truth the render can't
+show; see step 1b), + the rubric below. It is told NOTHING about which render is the skill's clean-room
+output vs the playground reference, and never reads any `.brep.ts`, the heal, the orchestration, or
 `apps/playground/**`. Structured return:
 
 - per label (`A`, `B`): `designed | partial | blob` + one-line reason (are the described features
-  present, legible, proportioned — or a lumpy primitive stack?),
+  present, legible, proportioned — or a lumpy primitive stack? do the measured body count + relations
+  match what the description implies?),
 - pairwise: `A-better | B-better | tie-good | tie-blob`,
 - `confidence`: `high | low`.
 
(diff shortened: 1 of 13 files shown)


andymai:
Thanks Greptile — verified both findings against the code; both are non-defects, declining with rationale:

**1. `manufacturability.ts:108` disposal** — `checkInterference` returns a **plain object**, not a WASM handle: `InterferenceResult` is `{ hasInterference: boolean, minDistance: number, pointOnShape1/2: Vec3 }` (`src/measurement/interferenceFns.ts:21`) — nothing to `using`-dispose. `validSolid(s)` returns `ok(s)` — the **same input solid rebranded** (`src/core/validityTypes.ts:171`), so disposing it would free the caller-owned body (a bug, not a fix). `measureVolume` returns `Result<number>`. This also matches the established pattern in the same function — `runChecks` already calls `validSolid`/`measureVolumeProps` per-body with no disposal (`checks.ts:64,80,105`) — and shape handles are FinalizationRegistry-backed (the #1172 disposal non-defect).

**2. `metrics.ts:27` bodyCount fallback** — `?? 1` is correct for every reachable state: `digestMetrics` returns early unless metrics exist, and `computeMetrics` always sets `manufacturability` whenever it sets `bodyRelations`, so the only path with `bodies` absent is a single solid → 1. The suggested `?? (manufacturability ? 1 : 0)` yields the same value for that case.

PR #1539, opened 2026-06-21
feat(brepjs-cad): decomposed, cross-checked design-judge rubric (Phase 1.5)

andymai:
## What

Phase 1.5 of the [design-judge upgrade](https://github.com/andymai/brepjs/pull/1531). Phase 1 gave the judge deterministic facts; this makes it **use them rigorously**. Both judges now **decompose the brief into its named features and grade each** (`present` / `correct`) before the overall verdict, reconciling counts against the Phase-1 measured facts.

Decomposition-first is the strongest lever against the two dominant LLM-judge failure modes: a vague "looks roughly right → pass", and hallucinating features that aren't actually rendered. No renderer or kernel changes.

## Changes

- **`judge.ts`**: `VERDICT` gains a `features[]` array (per-feature `present`/`correct`), placed **first** in the schema so the model decomposes before it concludes. `JUDGE_SYSTEM` directs: decompose → grade each → reconcile counts against the measured facts → `pass` only if every required feature is present and correct. New pure **`missingFeatures()`** helper.
- **`live.ts`**: folds the missing/wrong features into the judge `reason`, so the retry/heal feedback bundle gets actionable specifics (`[missing/wrong: motor bore]`) rather than one vague sentence — directly improving the heal loop's next-attempt hint.
- **`blind-judge.md`** (the $0 sub-agent path the heal/eval loop uses): the rubric now decomposes per-feature and reconciles with each render's measured-facts line as the basis for `designed`/`partial`/`blob`.

## Verification

- `tests/manufacturability.test.ts`: `missingFeatures` cases (absent + wrong-form flagged; all-correct → empty).
- Full package suite **166/166**; typecheck/eslint/prettier/knip clean.
- The LLM judge's behavior itself isn't unit-testable; the structured `features[]` schema is the reasoning scaffold, and the pure helper + plumbing are covered.

### DIFF
--- a/packages/brepjs-cad/bench/blind-judge.md
+++ b/packages/brepjs-cad/bench/blind-judge.md
@@ -28,10 +28,15 @@ output vs the playground reference, and never reads any `.brep.ts`, the heal, th
 - `confidence`: `high | low`.
 
 **Rubric (put in the judge prompt):** grade each render against the **description** as the absolute
-bar — the named features present and legible, the thing reads as manufactured/designed. Use the
-other render only to calibrate _how polished is achievable_, never to reward mere similarity. A part
-that differs from the other render but realizes the description as a designed object is `designed`.
-Ignore color, lighting, camera, and exact dimensions you can't read from the image.
+bar. **First decompose** the description into its named features (the holes, bores, walls, bodies,
+blades, teeth, slots it implies); for each render, check each feature `present` (legible) and
+`correct` (right count/form/proportion), **reconciling counts against that render's measured-facts
+line** (if the facts say N bodies, confirm you see N; treat a reported interference as intended vs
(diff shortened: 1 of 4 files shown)


andymai:
Thanks — considered the `.refine()` but declining: it would make the verdict more fragile, not safer.

The inconsistent combo (`present:false, correct:true`) is already handled correctly downstream — `missingFeatures` uses `!f.present || !f.correct`, so an absent feature counts as missing regardless of `correct`. Adding a `.refine()` to the zod schema means a single benign per-feature inconsistency fails the **whole** `messages.parse` structured-output validation, dropping the entire verdict to the "no parseable verdict" fallback (`pass:false, features:[]`). Losing a full verdict over one nonsensical flag is worse than tolerating it. The `correct` field's description already scopes it to a present feature's form/count.

PR #1540, opened 2026-06-21
feat(brepjs-cad): xray internal-reveal shot for the design judge (Phase 2a)

andymai:
## What

Phase 2a of the design-judge upgrade — the first **renderer-touching** layer. The judge rendered only opaque exterior views, blind to the internal features that ~88% of the mechanical corpus has (bores, tapped threads, shelled walls, internal teeth). This adds an **xray pass** that shows them through a translucent body. No aiming needed — the highest-value internal-visibility win at the lowest cost (reuses the viewer's existing `xray` view mode; no `brepjs-viewer` change).

## Changes

- **`viewer/screenshotApi.ts`**: generalize the snapshot hook from view-only (`onViewChange`) to a scene channel (`onScene`) + add `window.__setScene({ view, viewMode })`. `__renderView` stays as a solid-view shim on the same channel (back-compat).
- **`viewer/main.tsx`**: bridge `onScene` → `setView` + `setViewMode`.
- **`snapshot/shoot.ts`**: a `Shot` recipe (`view` + `viewMode`); the default recipe now adds an **`iso-xray`** capture alongside the four views, applied via `__setScene`.
- **`judge.ts`**: widen the SDK judge's image set to include the xray + tell it to use the xray to confirm internal features.
- **`blind-judge.md`**: name `iso-xray` as the detail view for internal-feature parts.

## Render-verified

On a shelled, bored box, the solid `iso` hides the interior; the `iso-xray` clearly shows the through-bore and cavity walls:

| iso (solid) | iso-xray |
|---|---|
| internals hidden | through-bore + cavity walls visible |

## Verification

- New `__setScene` test case; existing `screenshotApi` test migrated `onViewChange`→`onScene`.
- Full package suite **167/167**; typecheck/eslint/prettier/knip/boundaries clean.

Follow-ups (later increments): Set-of-Marks (kernel-anchored feature labels), aimed section + back-iso (Phase 3).

### DIFF
--- a/packages/brepjs-cad/bench/blind-judge.md
+++ b/packages/brepjs-cad/bench/blind-judge.md
@@ -14,8 +14,10 @@ is _pairwise_ (author vs reference) and _blind_ (labels stripped), and costs not
 
 One judge subagent (general-purpose — it must `Read` the PNGs) per part. Its **entire input**: the
 part's NL `description` + two rendered images each — **iso + one detail view, the SAME view pair for
-both renders** so the A/B comparison is fair (pick the detail view — front/top/right — that best
-exposes the features the description names) — labeled only **A** and **B**, **a one-line "measured
+both renders** so the A/B comparison is fair (pick the detail view — front/top/right, or **`iso-xray`
+for any part with internal features** like bores/cavities/shelled walls, since it shows them through a
+translucent body — that best exposes the features the description names) — labeled only **A** and
+**B**, **a one-line "measured
 facts" string per render** (body count + interference relations — the ground truth the render can't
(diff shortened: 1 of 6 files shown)


andymai:
Declining the import suggestion: `shoot.ts` deliberately defines `ViewName` locally and imports nothing from `brepjs-viewer` — the snapshot driver talks to the viewer page through puppeteer with plain string values, so it intentionally avoids a compile-time coupling to the viewer package. The new `ViewMode` follows that same established local-type convention (importing only it, while `ViewName` stays local, would be inconsistent). The literal union is identical to the viewer's `ViewMode`.

PR #1542, opened 2026-06-21
feat(brepjs-cad): reliable internal-bore detection for the design judge (keystone)

andymai:
## What

Un-defers the internal-feature metric (deferred in Phase 1 because curvature-sign can't tell a bore from a boss). This is the **keystone**: it unblocks aimed sections and feature-anchored Set-of-Marks (both need bore axes), and it enriches the judge's facts digest now (`internal bores: N, smallest radius R`).

## The reliable bore-vs-boss test (verified empirically)

A cylindrical face is an internal **bore** when **both**:
1. its outward material normal points *away from the axis* — `faceCenter` + orientation-adjusted `normalAt` vs the `faceAxis` radial. Probed: external cylinders ≈ **-1**, bore ≈ **+1** (clean ±0.99 split).
2. its angular extent is substantial — `uvBounds` uSpan **> 2.5 rad**. This excludes corner/edge fillets, which the normal test alone misclassifies (a convex fillet reads like a bore; but a quarter-fillet's uSpan ≈ π/2 fails the extent test, while a drilled bore sweeps ≈ 2π).

Both were established by probing the kernel directly, not assumed — curvature-sign was already proven insufficient.

## Changes

- **`report.ts`**: `BoreInfo { radius, axisOrigin, axisDir }`; `manufacturability` gains `bores[]` — the axes are carried so Phase 3 can aim a section cut / anchor a feature mark.
- **`manufacturability.ts`**: `computeCylinderFeatures` populates `minRadius` (smallest *substantial* cylinder — fillet strips excluded), `internalCylinderCount`, and `bores[]`.
- **`metrics.ts` / `judge.ts`**: the facts digest + judge prompt now carry the bore count + smallest radius (and note the xray view exposes them).

## Verification

- `tests/manufacturability.test.ts`: drilled box → **1 bore (r4, Z axis)**; filleted box → **0 bores, minRadius excluded**; external shaft → **minRadius set, not a bore**; digest surfaces the bore line.
- e2e via `brep verify --metrics --json` confirms the same on real parts.
- Full suite **171/171**; typecheck/eslint/prettier/knip/boundaries clean.

Unblocks: Phase 3 aimed sections (cut perpendicular to `bores[0].axisDir`) and feature-anchored marks (`H1..` at bore axes).


andymai:
Addressed Greptile:

**Fixed (1) bore radius conflation** — the digest's "internal bores" line now uses the smallest *bore* radius (`Math.min(...bores.map(b => b.radius))`), not the global `minRadius` (which includes external shafts). A bushing with a 5mm shaft + 10mm bore would previously have mislabeled the bore as 5mm.

**Fixed (3) duplicate count** — dropped `internalCylinderCount` from the stored report; the count is now derived on read as `bores.length` in `digestMetrics`, so the two can't drift.

**Declining (2) face-handle disposal** — consistent non-defect: `runChecks` already iterates `getFaces(shape)` undisposed (the topology-count block in `checks.ts`), the per-face calls here return plain values (Vec3 arrays / numbers / a plain `{origin,direction}` from `faceAxis`), and shape sub-handles are FinalizationRegistry-backed (the #1172 disposal non-defect precedent). Adding `using` only to this loop while the adjacent shipped code doesn't would be inconsistent.

PR #1543, opened 2026-06-21
feat(brepjs-cad): aimed section cut for the design judge (Phase 3)

andymai:
## What

Builds on the keystone bore detection (#1542): when a part has a bore, the judge now also gets a **section capture that cuts through the dominant bore axis**, opening it as a clean cross-section. The xray (Phase 2a) reveals internals translucently but is cluttered for *reading wall thickness* — an aimed cut is the clean read.

## How it aims (the keystone pays off)

`aimedSection(bores, bounds)` uses the bore axes from #1542: it picks the basis axis **most perpendicular** to the largest-radius bore and positions the plane to pass **through** the bore origin — so the plane *contains* the bore axis and the bore opens as a cross-section. Dominant-bore selection + clamped frac.

## Changes

- **`snapshot/aiming.ts`** (new, puppeteer-free so the CLI imports it on the default path): the pure aiming helper.
- **`viewer/screenshotApi.ts` / `main.tsx`**: `SceneControl` gains `section`; the bridge maps `frac` → absolute position via the model bounds and drives the existing section state (absent section → off, so non-section shots reset cleanly).
- **`shoot.ts`**: `Shot.section` + a `ShootOptions.section` that appends one aimed `section` capture to the default recipe.
- **`cli/main.ts`**: a snapshot now implies `--metrics` (so bores exist to aim from) and passes the aimed section to `shoot`.
- **`judge.ts` / `blind-judge.md`**: the judge is told about the section view and to use it for bore + wall-thickness reads.

## Render-verified

On a shelled, bored box, the `section` cut slices through the bore, showing the cavity walls + bore as a clean cross-section (reads wall thickness, where the xray is cluttered).

## Verification

- `tests/aiming.test.ts` (5 cases): perpendicular-to-Z-bore, never-along-the-bore, dominant-bore selection, frac-through-origin, out-of-bounds clamp.
- Full package suite **176/176**; typecheck/eslint/prettier/knip/boundaries clean.

### DIFF
--- a/packages/brepjs-cad/bench/blind-judge.md
+++ b/packages/brepjs-cad/bench/blind-judge.md
@@ -16,7 +16,8 @@ One judge subagent (general-purpose — it must `Read` the PNGs) per part. Its *
 part's NL `description` + two rendered images each — **iso + one detail view, the SAME view pair for
 both renders** so the A/B comparison is fair (pick the detail view — front/top/right, or **`iso-xray`
 for any part with internal features** like bores/cavities/shelled walls, since it shows them through a
-translucent body — that best exposes the features the description names) — labeled only **A** and
+translucent body — that best exposes the features the description names; for a bored part the `section` cut is the cleanest read of
+the bore + wall thickness) — labeled only **A** and
 **B**, **a one-line "measured
 facts" string per render** (body count + interference relations — the ground truth the render can't
 show; see step 1b), + the rubric below. It is told NOTHING about which render is the skill's clean-room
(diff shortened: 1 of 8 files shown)


andymai:
Verified both findings against renders:

**(2) section view angle** — good instinct, but render-tested it and the flat-on view regressed the shot: looking *along* the cut axis shows the **closed far half** (the clip keeps the side away from the camera, so its cut face points away), whereas the iso 3/4 angle looks *into* the opened cavity and legibly shows the bore + walls — more useful for the judge than a foreshortened-but-flat profile. (The suggested mapping was also inverted for this viewer: `VIEW_DIR` has `front` looking along Z and `top` along Y, so it'd be x→right, y→top, z→front — but moot given the above.) Keeping `iso`; added a comment explaining why.

**(1) duplicate `SectionSpec`** — declining, consistent with the codebase: `ViewName`/`ViewMode` are likewise duplicated as local string-literal types across the node CLI (`shoot.ts`/`aiming.ts`) and the browser viewer (`screenshotApi.ts`) on purpose, to keep the two build units decoupled (the browser bundle shouldn't compile-depend on `src/snapshot`, and `frac`-relative section is a snapshot concept, not a viewer-core one). It's a 2-field type; the duplication is the intended layer boundary.

PR #1545, opened 2026-06-21
feat(brepjs-cad): kernel-anchored Set-of-Marks for the design judge

andymai:
## What

Gives the judge **view-invariant, part-space addressing**: a marked iso capture labels each body `B1..` (at its bbox centroid) and each bore `H1..` (at its axis, from the #1542 bore detection). The same id lands on the same feature across every view because it's a 3D anchor projected per-view — so the judge can say *"B3's bore H2 is wrong"* unambiguously.

**Key feasibility point:** this delivers part-space addressing **without** per-body mesh color (the deferred-expensive MeshData multipart work) — marks need only feature centroids (kernel) + the camera matrix (drei `<Html>` projects them, rendered as DOM the screenshot captures).

## Changes

- **`snapshot/aiming.ts`**: `Mark` type + `featureMarks(bodies, bores)` — `B#` for multi-body parts (a single body needs no label), `H#` per bore.
- **viewer** (`screenshotApi.ts` / `main.tsx`): `SceneControl.marks`; a `<Marks>` component drawing drei `<Html>` badges at each 3D point.
- **`shoot.ts`**: `Shot.marks` + a dedicated `marks` iso capture appended to the recipe (plain views stay clean, the judge gets one labeled reference image).
- **`cli/main.ts`**: compute `featureMarks` from the report, pass to both snapshot paths.
- **`judge.ts` / `blind-judge.md`**: widen the image set to 8, tell the judge about the marked view and to cite `B#`/`H#` labels.

## Render-verified

| part | marks |
|---|---|
| 3-body compound | `B2`/`B3` badges at centroids (B1 stacks under B2 — coincident centroids in that fixture; distinct bodies render distinctly) |
| shelled bored box | `H1` at the bore axis; no `B#` (single body) |

## Verification

- `tests/aiming.test.ts` `featureMarks` cases: B# multi-body-only, H# per bore, single-body+no-bore → empty, combined B#+H#.
- Full package suite **180/180**; typecheck/eslint/prettier/knip/boundaries clean.


andymai:
Addressed Greptile:

**(1) B# label gap — fixed, but index-aligned rather than resequenced.** Switched body marks to `B<index>` (0-based) instead of `B<index+1>`. This matches how the measured-facts digest already refers to bodies (`"bodies 0&1: interfering"`), so the judge can cross-reference a mark with the facts — and it removes the gap by construction: the label IS the stable body index, so an unlocatable body (no bounds) simply has no mark, with no spurious renumber. I avoided the suggested *sequential* relabel because it would decouple the mark numbers from the report's body indices (B1 might be body 2), breaking the digest cross-reference. Added a test for the no-bounds-on-an-earlier-body case.

**(2) duplicate `Mark`/`SectionSpec`** — declining, consistent with #1543: these are deliberately duplicated as local types across the node CLI (`aiming.ts`/`shoot.ts`) and the browser viewer (`screenshotApi.ts`) to keep the two build units decoupled (the browser bundle shouldn't compile-depend on `src/snapshot`). Same call as the `ViewName`/`ViewMode` duplication.

PR #1547, opened 2026-06-21
fix(brepjs-cad): handle brepjs/playground refs + summarize body relations in the digest

andymai:
## What

Two fixes surfaced by a consolidated **end-to-end run** of the upgraded design judge on the recently-added playground mechanisms (scotch-yoke, worm-gear-drive, rack-and-pinion, three-jaw-chuck).

1. **`adaptReference` — `brepjs/playground` shim.** The newer examples import `color`/`present` from `brepjs/playground` (a playground-only runtime the CLI can't resolve), so they failed to render (`Cannot find package 'brepjs'`). Those helpers are cosmetic — `color(shape, hex)` tags a GLB material, `present(shape, artifacts)` attaches downloads — neither changes geometry, and CLI snapshots are grey anyway. Shim each to identity-on-the-shape so the geometry runs. Unblocks the recent mechanisms **and** the bim/sheet-metal corpus.
2. **metrics digest — summarize body relations.** `formatDigest` dumped every body pair (O(n²)); a 23-body part produced a **253-line wall** that buried the signal and would blow the judge's token budget. Now it lists only the interfering/touching pairs and summarizes the separates as a count (`(219 other body pair(s) sit apart)`).

## The e2e run that found these (all 4 judged with the full evidence set + facts digest; `usedMetrics: true` on every one)

| part | judge verdict | what the metrics caught |
|---|---|---|
| scotch-yoke | **pass** | confirmed exactly 3 bodies; crank pin engages slot (xray); only "distinct colors" failed (grey render, flagged as a render limitation) |
| worm-gear-drive | fail | **23 solids vs the required 3** — wheel/worm fragmented into dozens of 0.00mm-interfering bodies (backlash gap closed) — invisible in renders, caught via the digest |
| rack-and-pinion | fail | verified conjugate mesh via touching pairs at 0.00mm; failed "3 bodies" (kernel reports 6 — fragmentation) |
| three-jaw-chuck | fail | confirmed 5 bodies + 3 jaws gripping stock + Ø20 bore (10mm radius) the images left ambiguous; found missing scroll groove, serrations, 2 of 3 pinion sockets |

The upgraded judge **failed 3 of 4 references on real geometry/feature defects** the old exterior-only judge would have passed — which is the point. (These are also reference-quality issues worth fixing in the examples.)

## Verification

- `tests/adaptReference.test.ts`: playground-shim cases (color/present, aliases). `tests/manufacturability.test.ts`: digest lists interfering pairs, counts separates, doesn't dump them.
- Full package suite **183/183**; typecheck/eslint/prettier/knip clean.

### DIFF
--- a/packages/brepjs-cad/bench/adaptReference.ts
+++ b/packages/brepjs-cad/bench/adaptReference.ts
@@ -19,11 +19,37 @@ import { pathToFileURL } from 'node:url';
 // Render the result WITHOUT `--check`: playground `code` is type-checked against the looser Monaco
 // ambient surface, so strict-CLI type gaps (e.g. `sketchOnPlane('XY')`) are expected and irrelevant
 // to the image the judge needs.
+//
+// `brepjs/playground` (`color`, `present`) is a playground-only runtime the CLI can't resolve. Those
+// helpers are cosmetic — `color(shape, hex)` tags a GLB material, `present(shape, artifacts)` attaches
+// downloads — neither changes geometry, and CLI snapshots are grey anyway. Shim each to identity on
+// the shape so the part runs.
 
 const QUICK = 'brepjs/quick';
 
+/** Replace `import { a, b } from 'brepjs/playground'` with no-op shims returning the shape (1st arg). */
+function shimPlaygroundImports(code: string): string {
+  return code.replace(
+    /import\s*\{([^}]*)\}\s*from\s*['"]brepjs\/playground['"];?/g,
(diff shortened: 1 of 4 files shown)


andymai:
Thanks — declining the extra assertion: this test's purpose is the new summarize-vs-dump behavior (interfering pairs listed, separates counted and NOT dumped), which it covers; `distinct bodies` count is already exercised by the kernel-side metrics tests. Keeping it focused.

PR #1566, opened 2026-06-21
feat(brepjs-cad): add airfoils.md feature recipe (swept fans/props/impellers)

andymai:
## What

Closes the **airfoil-quality residual** the flywheel surfaced. The "Realize the designed object" discipline (#1563) got authors to *apply* `twistAngle`, but a twist on a flat paddle still reads as thin fins — this is a feature that needs a **recipe**, not just a nudge.

Adds `references/airfoils.md` with a **verified** recipe for swept blades (fans, props, impellers, vanes):
- a **cambered** airfoil section (two `sagittaArcTo` arcs → a thin crescent, not a flat strip),
- sketched **perpendicular to the span** and extruded **radially** (`sketchOnPlane('YZ', [R_HUB,0,0])` → +X),
- with a **pitch twist** along the span.

It names the two real mistakes (flat-rectangle section; extruding *up the axis* instead of radially). Wired into the reference index + the twisted-impeller cue in the designed-object section.

## Before / after

| | result |
|---|---|
| before (twist on flat paddle) | thin warped fins — judged `partial` |
| after (airfoils.md recipe) | broad cambered pitched blades radiating to the shroud — a real axial impeller |

## Verification (end to end, visual — not self-report)

1. The recipe **builds `ok:true` verbatim** (extracted from the markdown, before and after prettier).
2. A **clean-room author** given only the brief + `airfoils.md` produced a convincing swept impeller **first try** — confirmed by **rendering it** (the flat-fin author had *also* claimed `builtToSpec:true`, which is exactly why the verification is visual).

Markdown-only (new reference + two SKILL.md pointers).

### DIFF
--- a/packages/brepjs-cad/skills/implement/SKILL.md
+++ b/packages/brepjs-cad/skills/implement/SKILL.md
@@ -29,7 +29,8 @@ drops the one feature that makes it itself.
 1. Decompose the brief into named features, then mark the **ONE defining feature** — the geometry without
    which it's a generic blob. `GT2 pulley` → the belt-tooth profile (a smooth groove is not a GT2 pulley);
    `fluted knob` → full-height flutes around the **whole** perimeter (scattered scallops are not flutes);
-   `twisted impeller` → the blade twist (`sketch.extrude(h, { twistAngle })`, not flat blades); `scroll chuck` →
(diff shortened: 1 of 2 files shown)


andymai:
Applied — verified `origin?: PointInput | number`, so the scalar `R_HUB` is a normal-offset and produces identical geometry (same bbox, builds ok:true); clearer for a copied recipe. Updated the Notes mention too.

Issue #1549, opened 2026-06-21
Reference example quality: weld fragmented bodies + add missing features (surfaced by the upgraded design judge)


andymai:
## Context

The upgraded `brepjs-cad` design judge (deterministic body/interference metrics → decomposed rubric → xray → bore detection → aimed section → Set-of-Marks; PRs #1531/#1539/#1540/#1542/#1543/#1545, adapter+digest #1547) was run end-to-end on the recently-added playground mechanisms. It **failed 3 of 4 references on real geometry/feature defects** the old exterior-only judge waved through. These are pre-existing reference-quality issues the upgrade *surfaces* — tracked here as tech-debt (deferred from the judge-upgrade work).

Verify a fix with: `tsx packages/brepjs-cad/bench/adaptReference.ts <dir> <id>` → `node packages/brepjs-cad/dist/cli/main.js verify <dir>/<id>.ref.brep.ts --metrics --json <out>` and check `bodies` count / `manufacturability`. Then re-render (`brep snapshot`) + regen the committed `.webp` thumbnail + clear the playground gates (Monaco types, `playgroundExamples.test`, `exampleThumbnails.test`).

### `worm-gear-drive` — fragmented into 23 solids (intended: 3 colored bodies)
- Each colored body (frame / worm / wheel) is multiple **genuinely-disconnected** solids — bolt-head bosses, bearing blocks, etc. with real mm-scale gaps (not coincident-face tolerance; touching parts already weld). Body breakdown: 23 solids, 46k→60 mm³.
- **Caveat:** some gaps are *intentional* (the frame bores shaft clearances "so it never touches the rotating shafts"). A fix must distinguish defect-gaps (parts that should attach: bolt heads/bosses) from intended clearances — careful per-gap judgment, not a mechanical weld.
- Also flagged: worm↔wheel mesh reads as fully closed rather than a visible backlash gap.

### `rack-and-pinion` — fragmented into 6 solids (intended: 3 colored bodies)
- Same loose/disconnected pattern. Bodies: 30616, 12021, 8844, 1098, 522, 77 mm³.

### `three-jaw-chuck` — missing features (additive geometry)
- Judge (using xray/section/marks + facts) confirmed the core mechanism (5 bodies, 3 jaws gripping the Ø30 stock, Ø20 bore at 10mm radius) but found **genuinely missing**: the scroll-face spiral groove, the jaw serrations, and 2 of the 3 square pinion sockets (only 1 present).

## Suggested approach
[...]

PR #1567, opened 2026-06-21
fix(playground): replace 7 flawed examples with healed-skill author parts

andymai:
## What

The full-corpus eval-flywheel **re-run** (confirming a **+15–16 pt design lift** from the implement-skill heals: judge 44%→59%, mechanical 40%→56%, valid steady at 100%) blind-judged the healed skill's clean-room authors against the playground references. On **7 parts the fresh author `designed`-beat the reference** (which was `partial`/`blob`/invalid) — these references are the corpus's quality debt tracked in #1549.

This ports the author geometry in (geometry verbatim; reshaped to playground form: `brepjs/quick` imports, named-fn export, `compound([...])`→array for assemblies, `color()` kept for geneva).

| example | reference was | now |
|---|---|---|
| **gt2-pulley** | INVALID — 40 disconnected bodies | real toothed timing pulley |
| **geneva-drive** | a blob | clean slotted wheel + driver (colored) |
| **stud-brick** (`DEFAULT_CODE`) | loose un-welded studs | one welded interlocking brick |
| **modular-hose-segment** | partial | feature-complete |
| **vented-louvre-case** | partial | feature-complete |
| **din-rail-clip-enclosure** | partial | feature-complete |
| **finger-tab-split-joint** | partial | feature-complete |

## Gates
- `check:examples` — 59 snippets type-check (added `type Face` imports for two faceFinder predicates)
- `playgroundExamples.test` — **492 geometry tests pass across 4 kernels**
- `tsc -b` clean; all **7 thumbnails regenerated**

Fixed one ported defect: gt2's `return cut(...)` → `unwrap(cut(...))` — a `Result` default export passes `verify --check` (auto-unwrapped) but breaks the playground mesh path. Closes much of #1549.


andymai:
Addressed:
- **halfX (line 1913):** removed — it was a dead unused-var appeaser the author left in (the in-scope `halfX` for the vent honeycomb is a separate one, still used). No geometry change.
- **unsafe face-kiss fuse (line 1928):** keeping it. The part verifies (`ok:true`), meshes across all 4 kernels (playgroundExamples.test), and the blind judge rated it `designed`; `fuseAll({ unsafe: true })` is the documented assembly-weld pattern in `references/booleans.md`. Bumping the ridge 1mm would change verified geometry and force a thumbnail regen for a marginal robustness gain — not worth it here.

PR #1568, opened 2026-06-21
fix(brepjs-cad): heal 6 implement-skill findings from the full-flywheel re-run

andymai:
Six gaps surfaced by the lift-confirmation re-run, each RED→GREEN verified (re-author the failing part vs the patched skill → defect resolves, author quotes the new guidance):

- **unwrap the FINAL return** — a `Result` default export passes `verify` (auto-unwrapped) but renders nothing in a mesh/viewer path (gt2; caught the masked bug in #1567).
- **many-operand welds** — `fuseAll({unsafe:true})` leaves a loose N-solid compound for grilles/grids; fold with pairwise `fuse()` + confirm `getSolids===1` (fan-guard: 89→1).
- **derived extents** (the #1 first-try failure) — translated far edge = offset ± half-extent; overlap-widened face stays nominal; axis-redirected cylinder far end = base + axis·length (wall-mount, primitives, pipe-saddle, conduit).
- **gears.md** — GT2 rim as a single-valued polar `r(angle)` (a per-tooth absolute point list folds back → VALIDATION_FAILED).
- **threads.md** — fixed the un-unwrapped `closedWire(wire(...))` loft snippet (→ TS2345).
- **booleans.md** — a round body seated tangent on a flat plane is a non-manifold single-line contact; sink ~1mm.

Markdown-only.


andymai:
Applied — and it was the *root* of the gt2 bug: that line said a Result default export is 'supported', which is true for `verify` but false for rendering. Rewrote it to warn that `--check` green isn't enough — unwrap the final shape.

PR #1569, opened 2026-06-21
fix(playground): fix fragmented examples found by a verify --metrics audit

andymai:
## What

Audited **every** basics+mechanical example through `brep verify --metrics` (not just the mesh test — a fragmented compound still meshes, so `playgroundExamples.test` couldn't see this). Validity is now clean (0 `ok:False`), but the audit caught **4 examples that were valid-to-the-eye yet broken**:

| example | was | now |
|---|---|---|
| **gt2-pulley** | my prior port was valid but had dropped the hub + grub screw (and the *original* example was a **40-body INVALID** compound) | complete pulley — toothed rim, both flanges, hub stub + grub screw — as one **welded** solid |
| **fan-guard** | 141 loose solids (`fuseAll` unsafe didn't weld) | 1 solid (pairwise fuse) |
| **flanged-tee** | 9 loose solids | 1 solid |
| **axial-fan** | 20 fragmented bodies, flat blades | proper swept-airfoil impeller + shroud, a clean 2-body assembly (each a single solid) |

## Regression guard

Adds the 4 to `CONNECTED_BODY_EXAMPLES` so the `getSolids()===1` connectivity gate covers them. The mesh-only test passes on a fragmented compound; this is the explicit assertion that catches it. **508 geometry tests pass** (4 kernels), `check:examples` clean, 4 thumbnails regenerated.

(Follow-up worth doing: extend the connectivity/validity gate corpus-wide so no single-body example can silently fragment again.)


andymai:
Removed — dead import from the ported author code (no geometry change, thumbnail unaffected). check:examples still clean.

PR #1572, opened 2026-06-21
fix(playground): corpus-wide connectivity gate + fix 8 render-audit defects

andymai:
## What

A **render + design-judge audit of every example** (the visual half of the `verify --metrics` audit) found 8 examples that passed `eval+mesh` but were visibly/structurally wrong:

| Example | Defect | Fix |
|---|---|---|
| stud-brick | (my prior port) horizontal studs, can't interlock | reverted to correct top-stud design + **welded** (original left 9 unfused solids) |
| vented-louvre-case | (my prior port) floor fragmented into **55 solids** | reverted to clean original |
| bench-vise | T-handle clipped the body; nothing to grip | handle clear + **workpiece clamped** between jaws |
| rotary-potentiometer | missing 3 rear solder lugs | added |
| universal-joint | cross-spider floated free of the yokes | trunnions **engage** both yoke bores |
| wall-mount-flanged-junction-box | missing mounting ears | ears with obround slots + gussets |
| rack-and-pinion | fragmented bodies | each part one welded solid |
| tripod-rc2-plate | stray protruding tab | removed |

Each is render-verified and one connected solid per body.

## The gate (the durable win)

`playgroundExamples.test` now asserts **corpus-wide** that every returned body is one connected solid (was a curated 6-example list). `eval+mesh` passes on a fragmented compound — so this is what catches a failed weld / detached part before STEP/GLB export. **700/700 across 4 kernels**; `check:examples` + `tsc -b` clean; 8 thumbnails regenerated.

`★ Insight` — two of the 8 were my own #1567 port regressions: the design judge approved author versions that were geometrically broken (wrong stud orientation; 55-fragment floor). The judge rates *design*, not *connectivity* — so the corpus-wide gate is exactly the backstop that makes judge-driven porting safe.

(Solid validity is enforced separately by `verify`; this gate is connectivity-only since that's the eval+mesh blind spot. The axial-fan airfoil impeller is verify-valid + renders/meshes but sits near the strict-`isValidSolid` boundary — noted, not blocking.)


andymai:
Declining — the gate still correctly fails on a non-solid export, and `getSolids=0` in the message already signals 'this isn't a solid'. Not worth re-running the full shards for a diagnostic-wording tweak.

PR #1577, opened 2026-06-22
fix(playground): redesign bench-vise as a realistic box-way machine vise

andymai:
## Summary

Reworks the **Machinist Bench Vise** playground example from a toy-ish assembly into a realistic, mechanically-correct machine vise, and adds the small library capability it needed.

### Library: Acme/trapezoidal thread profile (`feat(operations)`)
`thread()` only built a sharp 60° V tooth. Power screws (vises, clamps) use a flat-crested trapezoidal profile, so added an optional `crest` parameter (axial half-width of a flat crest): `0` = V (default), positive = trapezoidal (`~0.18*pitch` ≈ Acme, near `toothHalfWidth` ≈ square). Validated + unit-tested.

### Example: box-way machine vise rebuild (`fix(playground)`)
- **Solid cast bed** (wide mounting flange + central slide bar) replacing the two thin rails. The movable jaw now **saddles the bar and slides on it** — a real box way.
- **Captured lead screw, crank at the front**, threading through a nut in the fixed jaw, with thrust + retaining collars at both ends — the jaw genuinely opens/closes and there's no floating screw stub.
- **Acme thread** (the new `crest` option) instead of a sharp V.
- **Serrated jaw plates** lifted clear above the thread crest (no clipping); **four counterbored bolt holes** in the flange ears with open clearance for a bolt + driver.

### History (addresses earlier review feedback on the branch)
Floating screw stub → seated/captured; z-fighting on the bottom → no coplanar faces; buried grippers → exposed; hidden crank → front-mounted + posed; crank hand clearance → standoff neck; mounting holes → added with real clearance; thread clipping the grips → plates raised; "not enough thread to open" / "doesn't look human-designed" / "wrong thread type" → full box-way redesign with an Acme lead screw.

### Verification
- `thread` unit tests (incl. flat-crest) checks out; `npm run check:examples`, `tsc -b`, lint + boundaries all pass
- `playgroundExamples` + `exampleThumbnails` for `bench-vise`: eval+mesh, every body a single connected solid, committed thumbnail (16 tests) pass
- Regenerated the brepjs ambient types (thread `crest`) and the thumbnail


andymai:
Addressed the revolve-angle docstring comment in a497c31c by reverting that line so the ambient diff is only the intended `crest` addition.

For the record it's a false positive on the units: the brepjs `revolve` JS API takes **radians** — its brepkit adapter (`src/kernel/brepkit/sweepOps.ts`) does `angle * (180 / Math.PI)` before the WASM, and the public default is `2 * Math.PI`. So the source docstring (radians) is correct; the committed ambient was just stale at "degrees". I left the ambient at its base value here rather than fix that staleness, since it's unrelated to this PR.

PR #1584, opened 2026-06-23
chore(bench): rerun cross-kernel comparison and add kernel-subset mode

andymai:
## What

Reruns the kernel-comparison benchmark suite across the three B-rep kernels (`occt`, `occt-wasm`, `brepkit`) and regenerates `benchmarks/results/latest.md` (was stale: occt-only, brepkit-wasm 2.112.0 → now all three, 2.115.8).

Adds a comma-separated `BENCH_KERNELS` mode to `initBenchKernels` (e.g. `BENCH_KERNELS=occt,brepkit,occt-wasm`) so a specific kernel subset runs in a single process — which is what populates the "vs occt" comparison column (the report keys its baseline to the native `occt` kernel).

## Why

- The previous `latest.md` only ever had occt-wasm rows with an empty comparison column, because the only multi-kernel modes (`all`/`both`) pull in **every** registered kernel — including `manifold`, a **mesh/CSG kernel that (a) doesn't belong in a B-rep comparison and (b) currently hangs this benchmark**. The subset mode lets the B-rep comparison run without it.

## Results (median, vs native occt)

brepkit wins every row; occt-wasm tracks native occt within ~±20% (expected for a WASM port). Note the harness times operations but does not verify output equivalence — brepkit's largest wins (intersect, fuse, chamfer) are analytic fast-paths; `cut(box,cyl)` (~1.2×) is the apples-to-apples row where both do full boolean work.

## Test plan

```
BENCH_KERNELS=occt,brepkit,occt-wasm npx vitest run --config vitest.bench.config.ts kernel-comparison
```
23 benchmarks pass; `latest.md` regenerated.

### DIFF
--- a/benchmarks/harness.ts
+++ b/benchmarks/harness.ts
@@ -266,14 +266,19 @@ export interface ReportSection {
 }
 
 function speedupText(baselineMedian: number, compareMedian: number): string {
+  if (compareMedian === 0 && baselineMedian === 0) return '≈ same';
   if (compareMedian === 0) return '**>100x faster**';
+  if (baselineMedian === 0) return '**>100x SLOWER**';
   const ratio = baselineMedian / compareMedian;
-  // Within rounding of parity — don't label measurement noise as faster/slower.
-  if (ratio.toFixed(1) === '1.0') return '≈ same';
-  if (ratio > 1) {
+  if (ratio >= 1) {
+    // Within rounding of parity — don't flag measurement noise as a speedup.
+    if (ratio.toFixed(1) === '1.0') return '≈ same';
     return `**${ratio.toFixed(1)}x faster**`;
   }
+  // Check the displayed slowdown multiplier, not the raw ratio, so a real
+  // sub-10% slowdown isn't rounded away into "≈ same".
   const inverse = 1 / ratio;
+  if (inverse.toFixed(1) === '1.0') return '≈ same';
   return `**${inverse.toFixed(1)}x SLOWER**`;
 }
 
--- a/benchmarks/harness.ts
+++ b/benchmarks/harness.ts
@@ -326,7 +326,12 @@ export function generateReport(
 
     for (const [_baseName, group] of groups) {
       const occtResult = group.get('occt');
-      for (const kid of KERNEL_ORDER) {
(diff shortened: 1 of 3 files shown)


andymai:
Addressed the review findings in 01887233:
- **P1 (subset aborts early):** the comma-list path now wraps each `initKernel` in try/catch and skips unavailable optional kernels, matching `initAllKernels`.
- **P1 (empty subsets):** an empty/comma-only `BENCH_KERNELS` now throws instead of silently initializing nothing.
- **P2 (whitespace):** the whole value and the single-kernel path are trimmed.
- **P2 (false speed labels):** `speedupText` now renders within-rounding-of-parity results as `≈ same`; `latest.md` regenerated accordingly (occt-wasm-vs-occt rows now read `≈ same`).

andymai:
Fixed in 5657ab6d (both T-Rex-verified findings):
- **Slowdowns can hide:** the `≈ same` guard now checks each direction's *displayed* multiplier, so 100 vs 105.2 renders `1.1x SLOWER` (regenerated `latest.md` now shows the mesh rows as `1.1x SLOWER` rather than `≈ same`).
- **Zero baseline leaks:** added an explicit `baselineMedian === 0` branch (`>100x SLOWER`) so a zero occt baseline no longer prints `Infinityx SLOWER`.

andymai:
Fixed in 502b42c8 (the T-Rex 'Report drops selected kernels' finding):
- `generateReport` now emits known kernels in `KERNEL_ORDER`, then appends any other kernel that produced results, so a kernel selected via a subset (e.g. `manifold`) is never silently dropped from `latest.md`.
- a subset where every kernel fails to initialize now throws instead of producing an empty report at exit 0.

andymai:
Fixed in a28d58a4 (both round-4 failure modes):
- **Guard bypass via duplicate/early-return:** ids are now de-duplicated, and the empty-subset guard checks `getAvailableKernels()` (actual registration) instead of a 'did not throw' counter — `initKernel` returns early after a failed first attempt, so the counter could read a false success.
- **Missing baseline:** a multi-kernel subset that omits `occt` now logs a warning that the report's `vs occt` column will be empty.

PR #1597, opened 2026-06-23
refactor(tech-debt): type occtWasm 2D curve casts as Curve2dObj

andymai:
## Problem

`c2d()` already returns the typed `Curve2dObj` discriminated union, yet ~7 call sites in `occtWasm/kernel2dOps.ts` did `c2d(curve) as any` — gratuitously throwing the type away and forcing every downstream `.__bk2d` / property access through `any`.

## Change

- Drop the gratuitous `c2d(curve) as any` casts (and the `basis: any` helper param), letting the `__bk2d` switches narrow `Curve2dObj` properly.

Tightening the type **surfaced two latent issues the `any` had hidden** — both addressed:
- `decomposeBSpline2dToBeziers` indexed the knot/multiplicity arrays without `noUncheckedIndexedAccess` guards → added `?? 0` (the index is provably in-bounds).
- Dead defensive code the non-optional union types prove unreachable — `&& basis` guards, `sense !== false`, `xDirAngle ?? 0`, redundant `Array.isArray`/assertions — all removed.

The genuinely-opaque bridges (transform/axis/bbox handles, the ellipse-arc extension object) keep their **sanctioned** `as any`.

## Verification

- `npm run typecheck` / `lint` / `validate` all green
- Full 2D suite (`2d`, `curve2dFns`, `blueprintFns`, `cannedSketches`, `make2dOffset`, `boolean2dRegression`) passes on occt-wasm — behavior identical

## Acceptance

- [x] No gratuitous `as any` on `c2d()` results; types narrow via `__bk2d`
- [x] Latent index-safety + dead-code issues the `any` masked are fixed/removed
- [x] `npm run validate` passes; no behavioral change

### DIFF
--- a/src/kernel/occtWasm/kernel2dOps.ts
+++ b/src/kernel/occtWasm/kernel2dOps.ts
@@ -258,9 +258,8 @@ export function getCurve2dBounds(curve: Curve2dHandle): { first: number; last: n
 }
 
 export function getCurve2dType(curve: Curve2dHandle): string {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- inspect trimmed basis
-  let cu = c2d(curve) as any;
-  while (cu.__bk2d === 'trimmed' && cu.basis) {
+  let cu = c2d(curve);
+  while (cu.__bk2d === 'trimmed') {
     cu = cu.basis;
   }
   return ow2d.curveTypeName(cu);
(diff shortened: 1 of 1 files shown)


andymai:
Verified these are not reachable at runtime, so the removed fallbacks are dead code, not safety nets:

- **Serialized/deserialized handles keep every field.** `serializeCurve2d` is `JSON.stringify(c)` and `deserializeCurve2d` is `JSON.parse(data)` — a total round-trip. Empirically confirmed: a round-tripped `Circle2d` still has `sense` and a round-tripped `Ellipse2d` still has `xDirAngle` (probe passed). So `sense !== false` → `sense` and `xDirAngle ?? 0` → `xDirAngle` can't surface `undefined`.
- **Every `Curve2dObj` is built by a constructor that sets all fields** (`makeCircle2d` always sets `sense`, `makeEllipse2d` always sets `xDirAngle`, the arc builders always set `basis`), and the union types mark them non-optional. `addCurveToBBox` already dereferences `c.basis` on trimmed curves without a guard, so the `&& basis` loop guards were already redundant with existing code.

The one place I *added* a guard is the genuine `noUncheckedIndexedAccess` site in `decomposeBSpline2dToBeziers` (`?? 0` on the knot/multiplicity arrays). Restoring the other guards isn't possible without re-introducing the `no-unnecessary-condition` lint errors that motivated this cleanup. Behavior is identical — full 2D suite passes on occt-wasm.

Issue #959, opened 2026-05-09
Error: The requested module 'blob:https://www.brepjs.dev/79eba15a-fc90-4c29-b33e-cb94798f4fc6' does not provide an export named 'Sketcher'


shendrick:
### Description

Here's an error I get locally and also when trying to use the Sketcher in the editor from the example.

https://www.brepjs.dev/tasks/sketching#the-sketcher-builder

The requested module 'blob:https://www.brepjs.dev/79eba15a-fc90-4c29-b33e-cb94798f4fc6' does not provide an export named 'Sketcher'

### Steps to Reproduce

Open the sketcher example above in the editor.

### Expected Behavior

Should not show error and should show geometry.

### Actual Behavior

The requested module 'blob:https://www.brepjs.dev/79eba15a-fc90-4c29-b33e-cb94798f4fc6' does not provide an export named 'Sketcher'

### Environment

Brepjs editor.

PR #1600, opened 2026-06-23
test(shaperef): un-skip shapeRef tests for brepkit faithful evolution

andymai:
## Draft — gated on the brepkit-wasm release with shape-evolution

brepkit **#962** implements faithful shape-evolution (persistent face identity through the GFA builder). `fuseWithEvolution`/`cutWithEvolution` now report each result face's *true* input source instead of the normal+centroid geometry heuristic, so the shapeRef roles resolve to **`'exact'`** confidence through booleans (the `rolePropagation` test's whole point).

### What this changes
Un-skips the **8 boolean-evolution** shapeRef tests:
`shapeRef.evolutionFuse`, `shapeRef.deletedFace`, `shapeRef.splitEvolution`, and `shapeRefIntegration.{multiStepReplay, filletEvolution, cutEvolution, rolePropagation, multipleTrackedFaces}`.

The remaining **2 stay skipped** — they are *not* evolution gaps, and their reasons are corrected:
- `geometricFallback` (cylinder + fillet): `filletWithEvolution` uses the geometry heuristic — fillets aren't GFA booleans, so faithful provenance doesn't apply; a filleted cylinder resolves fewer than the expected ≥2 roles.
- `brokenRef` (sphere identity): brepkit represents a sphere with **2 faces** (vs 1), so `assignRoles` returns 2 — a primitive-topology difference.

### Verification
Built `brepkit-wasm` from brepkit `main` (with #962), swapped the nodejs glue + `.wasm` into `node_modules`, and ran both suites against the brepkit kernel:

```
Test Files  2 passed (2)
Tests  16 passed | 2 skipped (18)
```

### Why draft
The published `brepkit-wasm` (2.115.9) has the volume fix (#959) but **not** shape-evolution (#962). This lands green once:
1. brepkit cuts a release containing #962, and
2. this PR bumps the `brepkit-wasm` dep to it (no type re-sync needed — #962 changed no wasm bindings).

Until then the brepkit-project CI here runs against 2.115.8 and these 8 will (expectedly) fail.

Refs andymai/brepkit#863


andymai:
Good catch — confirmed empirically. On `brepkit-wasm@2.115.8` the un-skipped `shapeRef.test.ts > returns BrokenRef for deleted face` failed (resolved a deleted-face ref with `'exact'` confidence). Bumped to `2.116.0` (`4798d320`), which includes the faithful shape-evolution support; the brepkit project now passes all 8 un-skipped tests (16 passed / 2 skipped for the two documented non-evolution divergences). Full `test:brepkit` on the branch: 4294 passed, 3 failed (pre-existing sweep/extrude volume divergences, unrelated — andymai/brepkit#965, #966).

Issue #1606, opened 2026-06-24
perf: content/placement-invariant caching + geometry instancing


andymai:
Tracking issue for a set of performance improvements centered on **doing less kernel work**: cache geometry independently of placement, and instance repeated geometry instead of copying it.

> **Status: ✅ complete (closed as a tracker).** Every actionable item below shipped and is verified in `main`. The only unchecked item — the *(Larger) dirty-flag recompute DAG* — is deferred **for cause** (it's application infrastructure, not library infrastructure; see the closing comment). OCCT-wasm parallel meshing was investigated and correctly closed as a no-op on the single-threaded WASM build.

## Why now

Profiling our own usage and downstream consumers shows the highest-leverage wins aren't "faster booleans" — they're avoiding redundant materialization and tessellation. Consumers already hand-roll these patterns on top of brepjs (multi-level placement-invariant sub-caches; template-instancing — cache one cell and clone per position; batched booleans; worker pools) because the library doesn't offer them natively. These belong in the library so every consumer can delete its bespoke layer.

## What brepjs already has

- Content-addressed CSG cache (`src/csg/evaluate.ts`), Merkle structural hash (`src/csg/hash.ts`), free-param projection + granular invalidation, DAG sharing (one-materialization-N-reuses, `tests/csg/csgPerf.test.ts:64-82`).
- `linear`/`circular`/`gridPattern` with `sameFace` glue (`src/operations/patternFns.ts`).
- Mesh cache (`src/topology/meshCache.ts`) and a mature disposal system (`src/core/disposal.ts`).

## Top concrete wins

- [x] #1603 — Placement-invariant move via a kernel location-only transform (`locate`). **Done** — `locate` across all four adapters + CSG `Translate`/`Rotate` re-tag instead of deep-copy (#1633); active on the default kernel via occt-wasm 3.6.0 `located()` (#1637). The mesh half landed too: translation-stripped reuse (#1649) and rotation/rigid-motion-stripped reuse (#1653) — a pure move/rotate no longer re-meshes. _Only the literal "one cache entry per geometry" collapse stays deferred (needs reworking borrowed-shape ownership); its narrow value is already covered by DAG sharing._
- [x] #1604 — Transform-only instancing node `{ sourceHandle, transforms[] }`: one shape × N placements; materialize on export. **Done** — #1613 (topology) + #1616 (CSG IR `Instance` node); released in brepjs 18.102.0.
[...]

PR #1607, opened 2026-06-24
feat(csg): add opt-in LRU eviction to the evaluator cache

andymai:
## What

Adds **opt-in LRU eviction** to the CSG `Evaluator` content cache via a new `maxCacheEntries` option. Addresses the cache-eviction item in the perf tracking epic #1606.

Previously the evaluator cache was an unbounded `Map` — entries lived for the evaluator's whole lifetime, so long-lived evaluators grew without limit.

## How

- New `EvaluatorOptions.maxCacheEntries` (positive integer; validated in the constructor). **Default `undefined` = today's unbounded behavior**, so this is fully additive and non-breaking.
- When set, least-recently-used entries are evicted **only between top-level `evaluate()` calls** — never mid-recursion, so operands still in flight are never freed.
- Disposal is **refcounted**. A single kernel handle can back several cache keys because boolean/compound identity short-circuits forward a child handle up unchanged (`Fuse(Empty, b) → b`, `FuseAll([x]) → x` — see `evaluators/booleans.ts`). A handle is disposed only when its **last** referencing entry is evicted, so eviction can never produce a use-after-free.
- The cache now **owns disposal directly**, replacing the `DisposalScope` (which couldn't release one handle on eviction without releasing the rest). The `registered` WeakSet is subsumed by the refcount map.
- Adds an `evictions` counter to `CacheStats`.

## ⚠️ Contract note (please eyeball)

The borrowed-shape lifetime contract is **unchanged by default**. When `maxCacheEntries` is set, it narrows: a returned shape is guaranteed valid only **until the next `evaluate()` call** (an older entry may be evicted/freed). This is documented on `EvaluatorOptions.maxCacheEntries` and in the file header. Callers who opt in must consume/copy results before the next call. Existing callers (no option) see no change.

## Tests

`tests/csg/cacheEviction.test.ts` — 8 tests, green across all kernel projects (occt-wasm, brepkit, manifold):

- unbounded default never evicts;
- bound caps entries + counts evictions; LRU keeps recently-used and evicts the stale one;
- **shared-handle safety**: `Fuse(b, Empty)` shares `b`'s handle; evicting one key keeps the handle alive (refcount), and the result stays measurable (would throw on a disposed handle if refcounting were wrong);
- **disposal accounting**: an evicted handle is freed and there's no leak after the evaluator is disposed;
- re-materializes correct geometry after eviction; `withEvaluator` passthrough; rejects non-positive-integer bounds.

Updated one strict `toEqual` in `csgEvaluator.test.ts` for the new `evictions` field.

---
🤖 Part of an autonomous performance pass against the #1606 epic.

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -1,5 +1,5 @@
 [
-  { "name": "total (all JS)", "path": "dist/*.js", "limit": "325 kB" },
+  { "name": "total (all JS)", "path": "dist/*.js", "limit": "330 kB" },
   { "name": "brepjs", "path": "dist/brepjs.js", "limit": "55 kB" },
   { "name": "brepjs/core", "path": "dist/core.js", "limit": "2 kB" },
   { "name": "brepjs/result", "path": "dist/result.js", "limit": "1 kB" },
--- a/src/csg/evaluate.ts
+++ b/src/csg/evaluate.ts
@@ -53,7 +53,8 @@ export interface EvaluatorOptions {
    * (a handle shared by several entries is freed only when its last entry is
    * evicted). Defaults to unbounded — entries live for the Evaluator's
    * lifetime. With a bound set, a returned shape is only guaranteed valid
-   * until the next successful evaluate() call. Must be a positive integer.
+   * until the next successful evaluate() call; a failed evaluate() is
+   * transactional — it leaves the cache unchanged. Must be a positive integer.
    */
   readonly maxCacheEntries?: number | undefined;
 }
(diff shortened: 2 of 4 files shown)


andymai:
Addressed the P2 (error-path eviction) in `2c8cc296`: `trimCache()` now runs only after `result.ok`, so a failed `evaluate()` can no longer evict and dispose a previously-returned good result. Tightened the contract docs to "next *successful* evaluate()" and added a regression test (an unbound-param failure must not evict a prior result).

andymai:
Addressed both P1s in `6f7564b5`:
- **Failed-insert growth**: a failed `evaluate()` now rolls back its own inserts (tracked in `pendingKeys`), so the bound holds under repeated invalid inputs and older results are untouched. Added a repeated-failure bound test.
- **Reentrancy**: trim/rollback now runs only when leaving the outermost `evaluate()` (gated on an `evalDepth` counter), so an `onStep` callback that re-enters `evaluate()` can't dispose the outer evaluation's in-flight operands. Added an onStep-reentrancy test.

andymai:
Addressed in `44ca5a3e`: before the outermost trim, the root key is now touched back to MRU, so a reentrant `onStep` that inserts entries after the root can't displace it from MRU and cause the returned shape to be evicted. Added a test that re-enters from the **root node's** onStep. With depth-gating + transactional rollback + this root-touch, `onStep` reentrancy on a bounded evaluator is fully covered.

andymai:
Reworked the design in `8a5e8de8` to collapse the reentrancy surface (findings #3–#6) rather than patch each corner: when `maxCacheEntries` is set, `evaluate()` is now **non-reentrant** (calling it from `onStep` throws), and reconciliation is a simple **commit-on-success / rollback-on-any-non-success** (including thrown errors) at a single non-nested level. This removes the depth counter, the root-touch, and the cross-call `pendingKeys` hazard entirely. Unbounded evaluators are unchanged. Also bumped the aggregate `total (all JS)` size budget 325→330 kB (the lib was exactly at the 325 kB ceiling; the opt-in feature added ~71 B brotlied).

Issue #1604, opened 2026-06-24
perf(operations): transform-only instancing node (one shape × N placements)


andymai:
## Problem

Repeated placement currently materializes N kernel copies: `linearPattern`/`circularPattern`/`gridPattern` (`src/operations/patternFns.ts`) call kernel bulk-copy then `fuseAll` (with `sameFace` glue). There is no representation for "one shape instanced at N transforms" that defers — or avoids — per-instance kernel materialization.

## Proposed change

Introduce an instancing handle/node — conceptually `{ sourceHandle, transforms: Mat4[] }` — that:

- holds **one** branded shape handle plus a list of placements (no per-instance kernel copy),
- materializes real transformed copies only when required (export/STEP, or when an individual instance is edited — a copy-on-write boundary),
- composes with the disposal system (`src/core/disposal.ts`) so N instances share one underlying kernel resource, disposed once,
- can emit a render payload of **one mesh + N matrices** (consumed by the viewer InstancedMesh path).

Keep `patternFns` as the "give me a fused solid now" path; add the instancing node as the "keep it cheap / array many" path.

## Acceptance

- Arraying a shape at N placements allocates one kernel handle, not N (assert via `getDisposalStats()`).
- Materialize-on-export produces geometry identical to the eager `gridPattern` path.
- Disposing the instance node frees exactly one underlying resource.

## Notes

Pairs with placement-invariant caching and the viewer InstancedMesh path (see the tracking epic). Downstream consumers build template-clone arrays by hand today because the library has no instancing primitive.

Effort: M–L · Priority: P1

andymai:
## Design note — transform-only instancing (#1604)

**Goal:** hold ONE source shape + N placements, never copying the kernel shape; produce real copies — or a render payload — only on demand. A 10×10 grid becomes *1 solid + 100 transforms*, for both export (materialize) and preview (one mesh + N matrices → #1605). The just-merged benchmark (#1612) showed the fused-array path is ~20–25% faster via `gridPattern`'s nested fuse, which `materialize({fuse:true})` can route through.

### Type — new branded handle in `src/core/shapeTypes.ts`
```ts
InstancedShape<D extends Dimension>  // NOT a ShapeHandle (no .wrapped)
  source: ShapeHandle<D>             // one kernel shape, owned
  placements: ReadonlyArray<Mat4>    // N transforms (plain data)
  [Symbol.dispose]()                 // disposes `source`; placements are GC'd
```
[...]

Issue #1603, opened 2026-06-24
perf(csg): placement-invariant content cache (transform-stripped key)


andymai:
## Problem

The CSG evaluator's content cache keys on a 4-tuple `(structuralHash, kernelId, projectedEnvHash, toleranceHash)` (`src/csg/evaluate.ts:134-138`). Because position is baked into a node's structural hash when geometry is built through the eager API (e.g. `box(w, h, d, { at: [x, y, z] })`), **identical geometry placed at different positions produces distinct cache entries**, and a pure move/rotate misses the cache.

This is the caching half of geometry instancing. If a node's cache key were the geometry *without* its placement, then:

- a pure translate/rotate becomes a cache **hit** (just re-tag the transform), and
- N identical shapes at N positions collapse to **one** materialization.

The IR already supports one-materialization-N-reuses when a node is referenced under multiple transform parents (`tests/csg/csgPerf.test.ts:64-82`). The gap is that the eager construction path folds the transform into the cached node instead of keeping it as a separate parent.

## Proposed change

- Represent placement as a transform node *over* a shared child, and exclude the parent transform from the child's structural hash (canonicalize so a node's hash is placement-free).
- Store the transform alongside the cache entry; apply it cheaply on retrieval.
- Ensure primitive `{ at }` sugar (`box`/`cylinder`/…) lowers to `transform(translate, primitive())` in the IR rather than baking the offset into the primitive's hash.

## Acceptance

- A shape built at the origin and the same shape translated produce **one** cache entry.
- 100 identical primitives at 100 positions materialize once (extend `tests/csg/csgPerf.test.ts`).
- Existing CSG tests pass; evaluated geometry unchanged.

## Notes

Consumers hand-roll this today — they maintain intrinsic-keyed (position-independent) sub-caches so identical parts generate once. Landing it in the library lets them drop the bespoke caches. Pairs with the instancing-node and viewer-InstancedMesh issues in the tracking epic.

Effort: M · Priority: P1

andymai:
[...]

PR #1610, opened 2026-06-24
chore(size): raise total-JS budget + add per-entry budgets for untracked exports

andymai:
Proactive headroom (user-approved) for the in-flight #1606 performance pass. The library was sitting at its 325 kB `total (all JS)` ceiling; #1603 (cache key) and #1604 (instancing node) add library code that would otherwise trip the budget per PR. Current actual is ~325 kB; per-entry budgets (incl. `brepjs` 55 kB) are unchanged.

Part of #1606.

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -12,5 +12,13 @@
   { "name": "brepjs/measurement", "path": "dist/measurement.js", "limit": "500 B" },
   { "name": "brepjs/io", "path": "dist/io.js", "limit": "8 kB" },
   { "name": "brepjs/worker", "path": "dist/worker.js", "limit": "2 kB" },
-  { "name": "brepjs/quick", "path": "dist/quick.js", "limit": "500 B" }
+  { "name": "brepjs/quick", "path": "dist/quick.js", "limit": "500 B" },
+  { "name": "brepjs/text", "path": "dist/text.js", "limit": "500 B" },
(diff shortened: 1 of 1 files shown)


andymai:
Good catch — addressed in `73aeb3d8`. Added explicit per-entry size-limit rows for the four public exports that lacked them: `text`, `projection`, `shapeRef`, and `kernel/occtWasm/occtWasmAdapter` (measured 140/158/137/93 B brotlied; limits 500 B / 1 kB, matching the existing leaf-entry style). So the aggregate headroom can no longer hide a per-entry regression. (The heavy implementation lives in dynamically-imported chunks, which the aggregate `dist/*.js` gate still covers.)

andymai:
These limits are correct and pass CI (the `size` check is green). The four entries are thin **re-export shims** — measured **140/158/137/93 B brotlied** via `npx size-limit` — because the heavy code (font/blueprint/metrics for `text`, the OCCT op modules for the adapter, etc.) lives in **dynamically-imported chunks**, not the entry's synchronous graph. size-limit minifies + brotlis before measuring, so the unminified on-disk file size isn't what's gated, and the aggregate `dist/*.js` row still covers the chunks. The 500 B / 1 kB limits give ~3–5× headroom over actual and match the existing leaf-entry rows (`measurement` 500 B for 152 B, `quick` 500 B for 263 B). If a future build change inlines a chunk into an entry, the tight gate firing is the *intended* signal to re-budget deliberately — so I'm keeping them tight.

PR #1612, opened 2026-06-24
test(benchmarks): compare gridPattern vs hand-rolled array fuse

andymai:
Adds a repeatable A/B benchmark: brepjs `gridPattern` (kernel bulk-copy + `sameFace` fuseAll) vs the hand-rolled translate-each-cell + `fuseAll` array that consumers like gridfinity-layout-tool build by hand, across 4×4/6×6/8×8 grids and all kernels.

**Finding (occt-wasm, apples-to-apples — both produce a fused solid):** `gridPattern` is ~equal at 16 cells and **~20% faster at 36–64 cells**, and the gap grows with grid size. The win is the kernel **bulk-copy + nested row/column fuse**, not `sameFace` glue (which barely helps under occt-wasm). Under brepkit, `gridPattern` returns an unfused compound (~0.5 ms, near-free) — great for instanced previews, not a fused export solid.

Informs the #1604 instancing design and the gridfinity tool's array path. Part of #1606.

Also noted (not fixed here): `gridPattern` isn't re-exported from the top-level `src/index.ts`, unlike its siblings `linearPattern`/`circularPattern`.

### DIFF
--- a/benchmarks/gridPattern.bench.test.ts
+++ b/benchmarks/gridPattern.bench.test.ts
@@ -1,0 +1,89 @@
+// A/B: brepjs `gridPattern` (kernel bulk-copy + `sameFace` fuseAll) vs the
+// hand-rolled "translate each cell + fuseAll" array that consumers like
+// gridfinity-layout-tool build by hand. Informs whether the instancing path
+// (#1604) should route through gridPattern, and whether the gridfinity tool
+// should switch from its hand-rolled clone+fuseAll.
+//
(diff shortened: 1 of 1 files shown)


andymai:
Fixed in `c88f5841`: each timed op now runs inside a `DisposalScope` (matching `benchmarks/csg.bench.test.ts`), so per-iteration shapes — base cells, translated copies, and the fused/pattern results — are disposed and the WASM heap no longer grows across samples. The finding is unchanged and actually cleaner (RME drops to ≤1.2%): gridPattern stays ~25% faster than hand-rolled at 8×8 under occt-wasm.

Issue #1605, opened 2026-06-24
perf(viewer): instanced render path (one geometry + N matrices)


andymai:
## Problem

`packages/brepjs-viewer` (R3F/three.js) renders shapes as individual meshes (or a consumer merges them). For an array of identical shapes this duplicates vertex data and forces a re-upload/re-merge whenever a placement or color changes.

## Proposed change

Add an instanced render path: given one mesh + a list of per-instance transforms (and optional per-instance color), render via `THREE.InstancedMesh` (`setMatrixAt` / `setColorAt`, `instanceMatrix.needsUpdate`). Moving or recoloring an instance becomes a matrix/color write, not a geometry rebuild. Consume the instancing node's payload (one mesh + N matrices) directly — see the instancing-node issue in the tracking epic.

## Acceptance

- Rendering N identical shapes uploads one geometry + N matrices (one `InstancedMesh` per unique geometry), not N geometries.
- Moving one instance updates a single matrix without re-uploading geometry.
- Per-instance selection/highlight overrides are kept off the hot instanced path (a separate non-instanced layer handles the small selected set) so they don't defeat batching.

Effort: M · Priority: P1

andymai:
Done — implemented by #1615 (merged 2026-06-24, `packages/brepjs-viewer`).

`buildInstancedMesh(data, placements, opts)` renders one source mesh + N per-instance transforms as a single `THREE.InstancedMesh`, with `instanceMatrix()` bridging brepjs row-major 4×4s (from `instancedMesh().instances`) to `THREE.Matrix4`.

**Acceptance**

- ✅ **One geometry + N matrices.** A single `BufferGeometry`/material/`InstancedMesh` backs all placements — `tests/geometry.test.ts` asserts `mesh.count === 2` while `geometry.position.count === 3` (source meshed once). Bounds are recomputed over the instance matrices (`computeBoundingSphere`) so a grid translated far from the source isn't wrongly frustum-culled.
- ✅ **Move one instance = one matrix write.** The returned standard `InstancedMesh` supports `setMatrixAt(i, …)` + `instanceMatrix.needsUpdate` with no geometry rebuild.
- ◑ **Selection/highlight kept off the hot path.** The library ships the instanced building block; keeping the small selected set on a separate non-instanced layer is a viewer-app composition concern (the playground / `verify --serve` viewer layers it) rather than something `buildInstancedMesh` itself enforces.

[...]

PR #1613, opened 2026-06-24
feat(operations): transform-only instancing layer

andymai:
First part of #1604 (the topology layer; the CSG IR `Instance` node is a fast-follow).

An `InstancedShape` holds **one source shape + N placements** with no kernel copies until materialized — a 10×10 grid is 1 solid + 100 transforms.

- `instance(shape, Mat4[] | Vec3[])` / `instanceGrid(shape, {cols,rows,pitchX,pitchY})`
- `materialize(inst, {fuse?})` → a **Compound** of placed copies (default — a layout is a scene of placements), or a single **fused solid**; a grid instance fuses via the faster kernel `gridPattern` path (~20-25% win, per #1612)
- `instancedMesh(inst)` → source meshed **once** + N matrices (the three.js `InstancedMesh` payload for #1605)
- owns the single source handle (`Symbol.dispose`); intermediate copies are GC-reclaimed
- also surfaces `gridPattern` from the top-level index (it was only on its siblings `linearPattern`/`circularPattern`)

Design + sign-off: #1604. Tests in `tests/instanceFns.test.ts`. Part of #1606.


andymai:
All six addressed in `2b0f66b7`:
- **Disposal (failed/compound/fused copies)**: `materialize` now builds the placed copies inside a `withScopeResult` scope — they're disposed once the independent Compound/fused result is built, and a failed `applyMatrix` disposes the copies made so far. Verified `makeCompound`/`fuseAll` keep their own refcounted references, so disposing the inputs is safe; added a 'leaves no stray copy handles' test.
- **Reject empty**: `materialize` returns Err for an empty instance.
- **Validate grid counts**: `instanceGrid` throws RangeError on non-positive-integer cols/rows.
- **Deep-copy matrices**: the Matrix4x4 path now clones each row, so caller mutation can't change `inst.placements`.

andymai:
Real bug — fixed in `cd17a067`. `combine()` registered every placed copy for disposal, but `fuseAll([x])` short-circuits to return that same handle, so a single-placement fused `materialize` disposed the shape it returned. Dropped `withScopeResult`; now I dispose the transient copies **except one the result aliases** (`c !== result.value`). Added a single-placement-fuse regression test. (The earlier 6 findings re-surfaced as outdated/line-null comments — those are addressed in `2b0f66b7` and covered by tests.)

andymai:
Grid-count validation is already in place (`2b0f66b7`): `instanceGrid` throws `RangeError` for non-positive-integer cols/rows, covered by the 'rejects non-positive-integer counts' test. This re-surfaced as an outdated comment after the lines shifted.

PR #1615, opened 2026-06-24
feat(viewer): instanced-mesh render path for instanced shapes

andymai:
#1605 — render a brepjs `instancedMesh()` payload (one source mesh + N placements) as a single `THREE.InstancedMesh`, so a 10×10 grid is **one GPU draw** instead of 100 meshes.

- `buildInstancedMesh(data, placements, opts?)`: one geometry (meshed once) + `setMatrixAt` per instance, matching the Renderer's material conventions (color/metalness/roughness/emissive).
- `instanceMatrix(m)`: brepjs row-major 4×4 → `THREE.Matrix4`; `buildInstancedMesh` accepts `THREE.Matrix4` or brepjs 4×4 arrays.
- exported from the viewer index. Consumes the `instancedMesh()` payload added in #1613.

Tests in `packages/brepjs-viewer/tests/geometry.test.ts`. Completes the instancing render story (#1604/#1606).

### DIFF
--- a/packages/brepjs-viewer/src/geometry.ts
+++ b/packages/brepjs-viewer/src/geometry.ts
@@ -9,6 +9,66 @@ export function buildGeometry(data: MeshData): THREE.BufferGeometry {
   return geo;
 }
 
+/** A per-instance transform: a THREE.Matrix4 or a brepjs row-major 4x4. */
+export type InstancePlacement = THREE.Matrix4 | ReadonlyArray<ReadonlyArray<number>>;
+
+/**
+ * Convert a brepjs row-major 4x4 (`[[row0],[row1],[row2],[row3]]`, as returned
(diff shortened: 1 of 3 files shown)


andymai:
All three addressed in `d16bc731`:
- **Instance bounds**: now call `mesh.computeBoundingSphere()` after `setMatrixAt` so the InstancedMesh's culling/raycast bounds span all placements (a far-translated grid no longer culls). Added a bounds-coverage test.
- **Viewer modes**: material now mirrors `Renderer` — `viewMode` drives side/wireframe/transparent/opacity/depthWrite, plus `clippingPlanes` and polygon offset.
- **Payload clarity**: documented that `data` is viewer `MeshData` (convert a brepjs `ShapeMesh` the same way you do for `Renderer`); `placements` is `instancedMesh().instances`. (The viewer stays decoupled from brepjs, so it can't import `ShapeMesh` directly.)

PR #1616, opened 2026-06-24
feat(csg): add an Instance node to the CSG IR

andymai:
Completes #1604 — express an instanced array **inside a parametric CSG tree**.

`instance(source, placements, fuse?)` builds an `InstanceNode`. `evalInstance` materializes the source **once** (DAG-shared via the evaluator cache) and applies the placements through the topology instancing layer (#1613) — a Compound by default, a fused solid with `fuse: true`.

Threads the new node kind through every exhaustive switch the compiler enforces: dispatch, `outputKindOf`, serialize (`toJSON` + a validated `readInstance` round-trip), edit (`walk` + `childrenOf`), and optimize. Tests in `tests/csg/csgInstance.test.ts` (build, evaluate compound+fuse, DAG sharing, JSON round-trip, traversal, optimize). Part of #1606.


andymai:
All three addressed in `c1ec7487`:
- **Non-finite matrices**: `readInstance` now validates with `isNumber` (which checks `Number.isFinite`), so NaN/±Infinity in serialized placements are rejected at the trust boundary. Added a test.
- **Copy placements**: the builder deep-copies the matrices, so later caller mutation can't desync geometry from the pre-computed `structuralHash`. Added a mutation test.
- **Fused kind**: `outputKindOf` for `fuse:true` now returns `'Solid'` (a fuse yields one solid) rather than the source kind, which could mis-report for non-3D sources. Added a test.

PR #1637, opened 2026-06-24
chore(deps): bump occt-wasm to 3.6.0 for located() fast path

andymai:
## Summary

Bumps the `occt-wasm` dependency from **3.4.0 → 3.6.0** (devDependency exact pin + `peerDependencies` range `^3.4.0` → `^3.6.0`).

occt-wasm 3.6.0 ([andymai/occt-wasm#188](https://github.com/andymai/occt-wasm/pull/188)) adds `located(id, matrix)` — a cheap, location-only rigid move that re-tags a shape's `TopLoc_Location` instead of deep-copying topology (O(1) vs O(topology size); measured ~57× cheaper on a 200-face solid).

brepjs already feature-detects `located` and falls back to the copying `transform` when absent (#1633), so this bump is what flips the default kernel onto the fast path.

## Validation

- `npm install` → lockfile updated to `occt-wasm@3.6.0` (resolved + integrity).
- `tsc --noEmit` clean.
- `vitest run --project occt-wasm`: **4466 passed, 76 skipped, 0 failed**.

Diff is dep-only (`package.json` + `package-lock.json`).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "knip": "6.16.1",
         "lint-staged": "17.0.7",
         "lz-string": "1.5.0",
-        "occt-wasm": "3.4.0",
+        "occt-wasm": "3.6.0",
         "prettier": "3.8.4",
         "puppeteer": "25.1.0",
         "size-limit": "12.1.0",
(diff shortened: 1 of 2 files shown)


andymai:
Addressed the Greptile 3/5 concern: the lockfile is now **strictly the occt-wasm 3.4.0→3.6.0 bump**.

The earlier `packages/brepjs-cad` `0.33.0→0.32.0` delta was `npm install` reconciling a **pre-existing staleness on `main`** — main's lockfile already carried `brepjs-cad@0.33.0` while its source `packages/brepjs-cad/package.json` is still `0.32.0` (the `0.33.0` bump is staged in the open release PR #1611, which regenerates its own lockfile on merge). I've restored those out-of-scope hunks to base so this PR doesn't touch brepjs-cad. `npm ci --dry-run` passes.

PR #1653, opened 2026-06-25
perf(csg): rotation-stripped mesh reuse — no re-mesh on a pure rotate

### LINKED ISSUE #1603: perf(csg): placement-invariant content cache (transform-stripped key)

andymai:
## What

Generalizes the placement-stripped mesh fast path in the CSG evaluator from **translation-only** to **full rigid motions** (translate + rotate). A pure rotation — or any translate/rotate chain — now reuses the inner geometry's tessellation instead of re-meshing the relocated shape.

Closes #1603. Refs #1606 (last concrete sub-item of the perf epic).

## Why

`peelTranslate`/`translateMesh` (shipped in #1649) only recognized `Translate` chains, so a pure `Rotate` fell through and re-tessellated. The shape path already re-tags rotations cheaply via `locate` (#1633 — active on the default kernel since occt-wasm 3.6.0); this closes the matching *"no re-mesh on a pure move"* half for rotation.

## How

- **`peelTranslate` → `peelRigid`**: peels outer `Translate`/`Rotate` nodes, composing them into one rotation (quaternion) + translation. Composition is outer∘inner (post-multiply while peeling outward); a rotation about a pivot `c` is handled as `R·(p − c) + c`. Stops at the first non-rigid node (`Scale`/`Mirror`/boolean/…) or a free param.
- **`translateMesh` → `transformMeshRigid`**: vertices get the full motion, normals get the rotation only (a unit rotation preserves length → no renormalization). A pure translation keeps the identity quaternion and degenerates to the **exact** previous vertex-shift path (normals shared by reference) — zero behavior change for the translation case.
- Angle convention matched to the kernel: CSG `rotate` is in **degrees** (`advancedOps.ts` applies `·π/180`).
- Reuses the existing `@/utils/quaternion.js` helpers (Layer 0) — no new math.

## Tests

`tests/csg/csgTranslationMesh.test.ts` extended with rotation coverage:

- 90°-about-Z maps `(x,y,z)→(−y,x,z)` on vertices **and** normals (asserts a fresh normals array).
- Mixed `translate∘rotate` composition and off-origin-pivot rotation, both checked **analytically** against the base mesh (kernel-independent → keeps brepkit green).
- A kernel-comparison test (reused mesh == the kernel's own located-mesh) gated to the OCCT family, pinning the degree/axis/sign convention against real kernel output.
- Face-group re-keying onto the placed rotated shape.

**Gate (`occt-wasm`):** full CSG suite **200/200** green; typecheck, eslint, prettier, `check:patterns`, boundaries all clean. brepkit stays green; manifold shows the same pre-existing placement-stripped divergence as the existing translation tests (face-hash semantics; not part of the CI gate).

### DIFF
--- a/src/csg/evaluate.ts
+++ b/src/csg/evaluate.ts
@@ -242,6 +242,37 @@ function isIdentityQuat(q: Quat): boolean {
 }
 
 /**
+ * Write `src` (flat xyz) rigidly transformed by quaternion (qw,qx,qy,qz) then
+ * translation (tx,ty,tz) into `dst`. The quaternion-rotate math is inlined with
+ * scalar locals — no per-vertex array allocation — because this runs once per
+ * mesh vertex/normal and helper-allocated vector math regressed the gridfinity
+ * benchmark 17% (see `utils/vec3.ts`). Pass zero translation for normals.
+ */
+function rotateXyzBuffer(
+  dst: Float32Array,
+  src: Float32Array,
+  qw: number,
+  qx: number,
+  qy: number,
+  qz: number,
+  tx: number,
+  ty: number,
+  tz: number
+): void {
+  for (let i = 0; i < src.length; i += 3) {
+    const vx = src[i] ?? 0;
+    const vy = src[i + 1] ?? 0;
+    const vz = src[i + 2] ?? 0;
+    const cx = 2 * (qy * vz - qz * vy);
+    const cy = 2 * (qz * vx - qx * vz);
+    const cz = 2 * (qx * vy - qy * vx);
+    dst[i] = vx + qw * cx + (qy * cz - qz * cy) + tx;
+    dst[i + 1] = vy + qw * cy + (qz * cx - qx * cz) + ty;
+    dst[i + 2] = vz + qw * cz + (qx * cy - qy * cx) + tz;
+  }
+}
+
+/**
  * Apply a rigid motion to a mesh: vertices get the rotation then translation,
  * normals get the rotation only (a unit rotation preserves length, so no
  * renormalization). UVs and triangles are motion-invariant. A pure translation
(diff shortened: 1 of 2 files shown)


andymai:
Addressed both P2 findings in `1b6cb52b`:

**Per-vertex allocations in the rotation hot path** — `transformMeshRigid` no longer round-trips through `quatRotate` (two temp arrays per vertex/normal). The quaternion-rotate math is now inlined with scalar locals in an allocation-free `rotateXyzBuffer` helper (one call per buffer, not per vertex), matching the hot-path inline-math convention documented in `utils/vec3.ts` (the 17% gridfinity regression). The translate-only path was already allocation-free and is unchanged.

**Missing inverse-ordering coverage** — added a `Rotate(Translate(Box))` test. As you noted, that ordering hits `peelRigid`'s Translate branch with a non-identity accumulated rotation (`rv = quatRotate(rot, v)`), which the existing `Translate(Rotate(Box))` test never exercised. Verified it produces `f(p) = rot_z90(p + [5,0,0]) = (−by, bx+5, bz)`, checked analytically against the base mesh.

occt-wasm + brepkit green; the inlined math is numerically identical (all existing rotation assertions still pass).

PR #1718, opened 2026-06-25
docs: add a Get in touch page and contact form

andymai:
## What

Adds a lightweight **Get in touch** path to the site and README so people building on brepjs can reach me directly — for project work, support, or just a question.

- **`/support` page** — short intro + a contact form. Open to project work but framed as a general get-in-touch, not a services pitch.
- **`ContactForm.vue`** — Web3Forms-backed form (name · company · email · message). It's a Vue component that **prerenders to static HTML** and submits via **AJAX**, so it works with and without JS and needs no backend or secret in the repo (the Web3Forms access key is public by design and only authorizes sending to my inbox).
- **Nav** — adds a "Get in touch" link.
- **README** — a short Get in touch section above Contributing, linking to the page.

## Notes

- Styling uses existing `--vp-c-brand-*` / divider tokens, so it matches light and dark themes.
- The page falls back to the default `og.png`; a custom OG card can be added later via `scripts/gen-og-docs.mjs`.

## Verification

- `npm run docs:build` ✓ — component compiles, page prerenders, nav valid
- Prettier ✓
- Worth a manual smoke test after deploy: submit once and confirm it lands at hi@andymai.com.

<a href="https://cubic.dev/pr/andymai/brepjs/pull/1718?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,10 @@ The command exits non-zero unless the report is `ok` (valid **and** every declar
 
 - [Gridfinity Layout Tool](https://github.com/andymai/gridfinity-layout-tool): Web-based layout generator for Gridfinity storage systems
 
+## Get in touch
+
+I'm Andy Aragon — I build and maintain brepjs and the geometry kernels beneath it ([occt-wasm](https://github.com/andymai/occt-wasm), [brepkit](https://github.com/andymai/brepkit)). Building something on brepjs and want a hand, or just have a question? **[Get in touch →](https://brepjs.dev/support)**
+
 ## Contributing
 
 See [CONTRIBUTING.md](./CONTRIBUTING.md) for development setup and guidelines.
--- a/apps/docs/.vitepress/config.ts
+++ b/apps/docs/.vitepress/config.ts
@@ -251,6 +251,7 @@ export default withMermaid(
         { text: 'Authoring with AI', link: '/agent/overview' },
         { text: 'API Reference', link: 'https://andymai.github.io/brepjs/' },
         { text: 'Playground', link: '/playground', target: '_blank', rel: 'noopener noreferrer' },
+        { text: 'Get in touch', link: '/support' },
         {
           text: major,
           items: [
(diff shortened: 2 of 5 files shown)


andymai:
Addressed both issues from cubic's review in d4915970:

- **P1 (no-JS submit leaked data via GET):** valid bug. The form had no `action`/`method`, so without JS a native submit would have done a GET to the current page and put the entered name/email into the URL. Added `action="https://api.web3forms.com/submit" method="POST"` plus hidden `access_key`/`subject` inputs, so the no-JS path now does a real POST to Web3Forms. `@submit.prevent` still intercepts for the AJAX path when JS is on — genuine progressive enhancement now.
- **P3 (errors not announced):** added `role="alert"` on the error message and `role="status"` on the success state so assistive tech announces both.

andymai:
Thanks Greptile — addressed your three P2s in the latest commit:

- **Concurrent submits can duplicate** ✅ — added an early `return` when `status === 'sending'`, so a double-submit racing the disabled-button state can't fire two requests.
- **HTML error becomes network error** ✅ — `res.json()` now falls back to `{}` on a non-JSON body, and a failed submit reports the HTTP status ("Couldn't send (error N)…") instead of mislabeling a server/proxy error as a network error.
- **No-JS success leaves site** — intentionally leaving this as-is. For the (very rare) JS-disabled path, Web3Forms' own success page gives a clear "message received" confirmation; a bare `redirect` back to /support would bounce the user to a blank form with no confirmation, which is worse UX. Not worth a dedicated thank-you page for a near-zero-traffic path.

PR #1720, opened 2026-06-26
fix(brepkit): harden rotation/scale matrix builders against non-Vec3 axis/center

### LINKED ISSUE #1719: brepkit adapter: rotate() with explicit axis/center throws "center is not iterable"
## Summary

While benchmarking the **brepkit** kernel adapter, replicating a Y‑axis rotation of a cylinder through the top‑level `rotate(shape, angle, position, direction)` threw `center is not iterable`. The brepkit **Rust kernel is unaffected** — building the same rotated solid directly (via a primitive's `axis` option) produces the correct result — so this is a brepkit **adapter (TS)** bug, not a kernel bug.

## Symptom

`TypeError: center is not iterable`, thrown from `rotationMatrix` in `src/kernel/brepkit/helpers.ts` (~L215):

```ts
// If center is non-zero, conjugate: T(center) * R * T(-center)
const [cx, cy, cz] = center;   // throws when `center` is not iterable
```

## Where it surfaced

Replicating a `cyl ∪ cyl` case — rotate a z‑axis cylinder +90° about Y, then translate — through the brepkit adapter (occt‑vs‑brepkit head‑to‑head benchmark, local brepkit‑wasm 2.120.6). Worked around by constructing the second cylinder directly:

```ts
// instead of rotate(cylinder(3,20), 90, …, [0,1,0]) + translate:
const c2 = cylinder(3, 20, { at: [-10, 0, 0], axis: [1, 0, 0] });
```

## Analysis

- `rotationMatrix(angleDeg, axis = [0,0,1], center = [0,0,0])` — the `center` default **only guards `undefined`**. Any non‑iterable, non‑`undefined` value reaching `const [cx, cy, cz] = center` throws. (`scaleMatrix` just below has the same `const [cx, cy, cz] = center` pattern.)
- The happy‑path mapping looks correct: `rotate(shape, angle, position, direction)` (`src/topology/transformFns.ts`) → `rotateWithHistory(…, direction, position)` → adapter `(…, axis, center)` → `rotate(bk, shape, angleDeg, axis, center)` → `rotationMatrix(angle, axis, center)`. So the trigger is a **specific path**, most likely:
  - the **evolution replay** `composeOpsToMatrix` → `rotationMatrix(op.angle, op.axis, op.center)` (`src/kernel/brepkit/evolutionOps.ts` ~L891), where `op.center` is typed `… | undefined` and may arrive as a non‑array; or
  - a particular `rotate`/CSG‑builder call form that places a non‑Vec3 in the `center` slot.

I did not isolate a minimal one‑liner (the obvious `rotate(cyl, 90, [0,0,0], [0,1,0])` static‑traces as correct), so flagging the fragile site for confirmation rather than asserting an exact repro.

## Suggested fix

Make `rotationMatrix` (and `scaleMatrix`) robust to a non‑Vec3 `center`/`axis` before destructuring, e.g.:

```ts
const c = Array.isArray(center) && center.length === 3 ? center : [0, 0, 0];
const [cx, cy, cz] = c;
```

and confirm which call path delivers a non‑iterable `center` (likely the evolution‑replay op).

---

_Found during the brepkit analytic‑boolean perf validation; the kernel produced correct geometry via the `{ axis }` workaround, so impact is limited to the `rotate(...position, direction)` adapter path._

andymai:
## Summary

Fixes #1719 — a `rotate(...)` through the **brepkit adapter** could throw a cryptic `TypeError: center is not iterable` from `rotationMatrix` in `src/kernel/brepkit/helpers.ts`.

**Root cause:** `rotationMatrix` / `scaleMatrix` destructure their vector args directly (`const [cx, cy, cz] = center`) and rely on a default parameter (`center = [0,0,0]`) for the empty case. But TS default parameters only fire on `undefined` — any *other* non-iterable value reaching that slot (a `{x,y,z}`-shaped object, `null`, a scalar from a malformed compose op) sails past the default and throws on the destructure. `scaleMatrix` had the identical pattern.

## Fix

Added a small `asVec3(value, fallback)` coercion helper and routed `axis`/`center` through it in both builders. It:
- extends the *existing* `undefined → default` tolerance to **any** non-Vec3 input, so the builders never throw;
- validates each element is a finite number, so a malformed value can't silently inject `NaN` into the matrix.

Valid `Vec3` inputs are unchanged (same numbers out).

## Investigation note

The issue reporter couldn't isolate a minimal repro. I traced every `rotationMatrix`/`scaleMatrix` caller (public `rotate({at, axis})`, fluent `shape().rotateX()`, the CSG evaluator, and the `composeTransform` replay path) — all *static* paths deliver clean `Vec3`s. The crash only happens when an untyped value crosses the JS↔WASM boundary, where `no-unsafe-*` is disabled and the `Vec3` annotation isn't enforced at runtime. Hardening the fragile destructuring site is the right fix, as the issue recommended.

## Tests

New `tests/brepkitMatrixHelpers.test.ts` (pure-function, no WASM):
- correct rotation/scale matrices for valid input + valid non-origin center;
- non-iterable / `null` center and non-iterable axis no longer throw and fall back cleanly.

Placed in a dedicated file because the mock-adapter suite (`brepkitAdapter.test.ts`) is **excluded from every Vitest project** and never runs in any gate — a regression there would look covered while protecting nothing. This file runs in the default occt-wasm gate and all kernel projects.

**RED→GREEN verified:** stashing the fix reproduces the exact `center is not iterable` error; restoring it passes. Full pre-commit suite (3972 tests) green.

<a href="https://cubic.dev/pr/andymai/brepjs/pull/1720?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/kernel/brepkit/helpers.ts
+++ b/src/kernel/brepkit/helpers.ts
@@ -218,7 +218,11 @@ export function rotationMatrix(
   axisInput: readonly [number, number, number] = [0, 0, 1],
   centerInput: readonly [number, number, number] = [0, 0, 0]
 ): number[] {
-  const axis = asVec3(axisInput, [0, 0, 1]);
+  const rawAxis = asVec3(axisInput, [0, 0, 1]);
+  // A zero-length axis can't be normalised (len === 0 → division by zero → NaN);
+  // fall back to +Z, matching the `undefined`/non-Vec3 default.
+  const axis: readonly [number, number, number] =
+    rawAxis[0] === 0 && rawAxis[1] === 0 && rawAxis[2] === 0 ? [0, 0, 1] : rawAxis;
   const center = asVec3(centerInput, [0, 0, 0]);
   const rad = (angleDeg * Math.PI) / 180;
   const c = Math.cos(rad);
--- a/src/kernel/brepkit/helpers.ts
+++ b/src/kernel/brepkit/helpers.ts
@@ -186,12 +186,40 @@ export function translationMatrix(x: number, y: number, z: number): number[] {
   ];
 }
 
+/**
+ * Coerce a possibly-malformed value into a Vec3 tuple, falling back to `fallback`.
+ *
+ * The matrix builders sit at the JS↔WASM boundary where `no-unsafe-*` is off, so
+ * an `axis`/`center` typed as `Vec3` can arrive as a non-iterable at runtime (a
+ * `{x,y,z}` object, `null`, a scalar from a malformed compose op). Default
+ * parameters only guard `undefined`, so destructuring such a value throws a
+ * cryptic `center is not iterable`. This extends the existing `undefined → default`
+ * tolerance to any non-Vec3 input instead of crashing.
+ */
+function asVec3(
+  value: unknown,
+  fallback: readonly [number, number, number]
+): readonly [number, number, number] {
+  if (!Array.isArray(value) || value.length < 3) return fallback;
+  const [x, y, z] = value as [unknown, unknown, unknown];
+  return typeof x === 'number' &&
+    typeof y === 'number' &&
+    typeof z === 'number' &&
+    Number.isFinite(x) &&
+    Number.isFinite(y) &&
+    Number.isFinite(z)
+    ? [x, y, z]
+    : fallback;
+}
+
 /** Build a row-major 4x4 rotation matrix (angle in degrees, optional axis/center). */
 export function rotationMatrix(
   angleDeg: number,
-  axis: readonly [number, number, number] = [0, 0, 1],
-  center: readonly [number, number, number] = [0, 0, 0]
+  axisInput: readonly [number, number, number] = [0, 0, 1],
+  centerInput: readonly [number, number, number] = [0, 0, 0]
 ): number[] {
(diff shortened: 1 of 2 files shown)


andymai:
Addressed the cubic P2 in 9f8c380: a zero-length axis `[0,0,0]` is finite (so it passes `asVec3`) but can't be normalised — `len === 0` would produce `NaN` matrix entries. It now falls back to the default `+Z`, matching the non-Vec3/undefined behaviour, with a dedicated regression test.

Note on the failing `Vercel – brepjs-studio` check: it points at `rootDirectory: apps/studio`, which doesn't exist in this repo (only `apps/docs` and `apps/playground`), so that deploy fails on every PR independent of content. It's not a required status check (`ci-pass` is the only required check) and is unrelated to this change.

PR #1731, opened 2026-06-26
feat(brepjs-cad): grade judge quality against a reference exemplar

andymai:
## Problem

The live eval judge (`bench/judge.ts` → `bench/loop.ts`) returned an **absolute boolean with no comparand**, so the repair loop converged the instant a part cleared the "valid solid with the named features present" floor (`loop.ts`: `auto.pass && (judgePass ?? true)`). It had no way to rank *quality* above that floor.

A controlled A/B test (one part, five variants, judged on the real 5-view loop input) confirmed the failure:

| Variant | Reality | Old binary judge |
|---|---|---|
| golden | clean | **pass** |
| crude | all features, grotesque proportions | **pass** |
| scaled | 2× uniform | **pass** |
| thin-wall | 0.2 mm wall, near-degenerate | **pass** (`manufacturable:true`) |
| no-bore | named feature deleted | fail |

Everything that cleared the floor passed identically — no gradient. The enriched evidence layers (xray/section/marks) couldn't help because the verdict *type* couldn't express "worse," and a judge shown one part in isolation has nothing to calibrate against.

## Fix

Put a **reference exemplar + a graded relative verdict on the gating path**:

- **`judge.ts`** — `JudgeInput.referencePngPaths`; `VERDICT.quality` (`'worse' | 'on-par' | 'better'`, nullable). `JUDGE_SYSTEM` now grades quality *relative to a labelled reference* (explicitly **not** on size/scale) as the gradient above the absolute `pass` floor, and treats the image as decisive when the deterministic bore count over-reports.
- **`loop.ts`** — convergence now also requires `quality !== 'worse'`: a passing-but-worse part keeps iterating with a reframed "valid but lower quality" feedback bundle. An absent quality/reference does **not** block (fully back-compatible with the no-reference path).
- **`live.ts`** — render the corpus example once per prompt via the existing `adaptReferenceCode` and hand it to the judge. The reference is **free** — the playground *is* the corpus.
- **`playgroundCorpus.ts` / `prompts.ts`** — carry the example source as `EvalPrompt.referenceCode`.
- **`score.ts`** — `QualityGrade` type + `quality` on the attempt/result + scorecard surfacing (`q:worse` row tag, "quality vs ref" summary).

## Proof

Re-ran the same A/B with the reference shown:

| Variant | Old | New (vs reference) | Loop now |
|---|---|---|---|
| crude | pass → done | pass · **worse** | retries |
| scaled | pass → done | pass · **on-par** | converge (scale not penalized) |
| thin-wall | pass → done | pass · **worse** | retries — *the 0.2 mm wall is obvious side-by-side* |
| no-bore | fail | fail · worse | retries |

The gradient the binary lacked now exists. Notably the relative comparison catches the thin wall that no single-part absolute view could.

## Verification

`bench/` is outside the root pre-commit lint-staged/vitest scope, so it was verified with the package's own gates:

- `npm run typecheck -w brepjs-cad` ✅
- `npm run lint -w brepjs-cad` ✅
- `npm run test -w brepjs-cad` ✅ — 193 tests (+4 new loop gating tests covering: worse blocks convergence, the gradient is climbable, the retry framing, and back-compat for an absent grade)
- prettier ✅

## Caveat

The graded judge was validated **in-session** (the harness default `judgeModel` is `claude-opus-4-8`, so me-as-judge is the real decision conditions) because the billed `judge()` path needs an `ANTHROPIC_API_KEY` not present locally. The definitive confirmation is one billed `eval:live` run on CI, where the new `q:*` columns appear and the loop iterates past first-pass on low-quality parts instead of stopping at the floor.

<a href="https://cubic.dev/pr/andymai/brepjs/pull/1731?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/brepjs-cad/bench/judge.ts
+++ b/packages/brepjs-cad/bench/judge.ts
@@ -12,7 +12,9 @@ const JUDGE_SYSTEM = `You are a meticulous CAD reviewer. You are shown rendered 
 
 Judge ONLY whether the rendered geometry matches the request and rubric: the right features are present (holes, walls, grooves, fillets, slots), the overall shape is right, and proportions are roughly correct. Ignore color, lighting, camera, and exact dimensions you cannot measure from a render. Be strict about missing or wrong features — a bored hole that isn't there, a hollow part rendered solid, or the wrong overall form is a fail.
 
-You may also be given "Measured facts" computed deterministically from the kernel (exact body count, which bodies sit apart vs touch/overlap, and the count of internal bores with their smallest radius). Treat these as ground truth the render cannot show, and reconcile them with what you see — if the facts report N bodies or B internal bores, confirm you see them (the xray view exposes bores). A reported interference is ambiguous on its own: decide from the request and image whether it is an intended assembly (e.g. an exploded view or a part that legitimately overlaps another) or an accidental collision. Also judge manufacturability: whether the part reads as a producible object (no zero-thickness walls, stray disconnected bodies, or self-colliding geometry the request did not ask for).
+You may also be given "Measured facts" computed deterministically from the kernel (exact body count, which bodies sit apart vs touch/overlap, and the count of internal bores with their smallest radius). Treat these as ground truth the render cannot show, and reconcile them with what you see — if the facts report N bodies or B internal bores, confirm you see them (the xray view exposes bores). A reported interference is ambiguous on its own: decide from the request and image whether it is an intended assembly (e.g. an exploded view or a part that legitimately overlaps another) or an accidental collision. The bore count can over-report (a boss/seam cylinder is sometimes miscounted as a bore), so the IMAGE is decisive when the facts and what you see disagree about whether an internal hole exists. Also judge manufacturability: whether the part reads as a producible object (no zero-thickness walls, stray disconnected bodies, or self-colliding geometry the request did not ask for).
+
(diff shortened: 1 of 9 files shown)


andymai:
Thanks @cubic-dev-ai and @greptile-apps — addressed the review:

**Fixed (commit `3e278ba`):**
- **`score.ts` schema version** (cubic): bumped `SCHEMA_VERSION` 2 → 3 since the attempt/result shape gained `quality`.
- **`score.ts` quality-vs-ref summary scope** (cubic + greptile): restricted the summary to parts that cleared the floor — `r.auto.pass && r.judgePass === true` (the full floor, slightly stricter than the suggested `judgePass`-only), so a floor-failing part carrying a `'worse'` grade no longer pollutes an above-floor metric.
- **`loop.ts` feedback framing** (cubic + greptile P1): `onlyQuality` now requires `judgePass === true`, so a real judge *failure* keeps the corrective "did not pass" framing instead of being softened to a quality nudge. (Equivalent to greptile's `(judgePass ?? true)` here, since `quality === 'worse'` implies the judge ran.)
- **`judge.ts` no-reference guard** (cubic): when no reference is shown, the verdict's `quality` is forced to `null` so a spurious model grade can't block convergence — preserving the back-compat path.

Added 2 tests locking the floor-scoped summary and the judge-failure framing.

**`live.ts:160` — reference render path (cubic):** documented rather than changed, with evidence. I checked the alternatives empirically against an `ok:false` shape:
- `verify --check --step` (current): no STEP written.
- `export --step`: also no STEP (`written: []`).
- Only the in-process `snapshot`/`runPart` render produces a STEP for an invalid shape (it gates on `Result.Ok`, not `report.ok`) — but that path needs an in-process kernel the `eval:live` process doesn't initialize.

So a "lenient export" swap wouldn't help. In practice the playground corpus is gated to valid single-connected solids (the connectivity test in `playgroundExamples.test.ts`), so references build a STEP under the existing path; an unbuildable reference correctly falls back to absolute (`quality:null`) grading rather than anchoring against a broken exemplar. Added a comment explaining this (commit pending).

PR #1740, opened 2026-07-04
fix(origins): recover split coplanar face origins across boolean seams

andymai:
## Problem

The geometric origin-matching fallback (`findBestOriginMatch`) recovers face origins for boolean-regenerated faces by matching surface normal + centroid against input faces. It rejected any candidate whose centroid sat **>10mm away** (`distSq > 100`) or whose **normal was flipped** (`dot < 0.707`). Two real cases slipped through:

- **Split large planar face** — a boolean that splits a wide tagged floor (a cutout floor sliced by an overlapping deeper cutout) leaves pieces whose centroids drift 10–20mm from the parent's, past the near-match cutoff.
- **Cut normal flip** — a cut flips the cavity face's outward normal versus the tool face it came from, so the cavity floor is *antiparallel* to its origin's input face.

Both leave the surface with no origin → it falls back to body color in multi-color consumers.

**Downstream symptom:** gridfinity-layout-tool [#2443](https://github.com/andymai/gridfinity-layout-tool/issues/2443) — overlapping shadow-board cutouts of the same color printed part of a floor in the body color.

## Fix

Add a second acceptance path to `findBestOriginMatch`: a **coplanar match** — the output face lies on an input face's plane (orientation-agnostic normal via `|dot|`, perpendicular offset ≤ 0.1mm²), regardless of in-plane centroid distance.

- Legacy **near match** (signed normal, ≤10mm) is unchanged and always outranks coplanar matches, so existing behavior is preserved.
- Among coplanar candidates the nearest in-plane wins, and the perpendicular-offset gate cleanly separates parallel surfaces at different depths (e.g. a cutout floor vs the bin floor).

## Tests

`tests/originGeometricFallback.test.ts` gains two regression tests — a **fuse-split** face and the direct **cut-split** #2443 shape (block minus a wide shallow cavity minus an overlapping deeper cavity). Both fail on `main` (6 failures across occt/brepkit/opencascade) and pass with the fix. The experimental mesh-CSG (`manifold`) kernel attributes split faces through its own path; that divergence is registered in `kernelDivergences.ts` and the tests skip it there.

Verified end-to-end: building this branch and swapping it into gridfinity-layout-tool makes the #2443 reproduction's previously-untagged floor fully tagged (0 mm² untagged at the cutout depth).

<a href="https://cubic.dev/pr/andymai/brepjs/pull/1740?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.size-limit.json
+++ b/.size-limit.json
@@ -1,5 +1,5 @@
 [
-  { "name": "total (all JS)", "path": "dist/*.js", "limit": "340 kB" },
+  { "name": "total (all JS)", "path": "dist/*.js", "limit": "342 kB" },
   { "name": "brepjs", "path": "dist/brepjs.js", "limit": "57 kB" },
   { "name": "brepjs/core", "path": "dist/core.js", "limit": "2 kB" },
   { "name": "brepjs/result", "path": "dist/result.js", "limit": "1 kB" },
--- a/src/topology/metadata/originTrackingFns.ts
+++ b/src/topology/metadata/originTrackingFns.ts
@@ -102,7 +102,7 @@ export function propagateOriginsFromEvolution(
 /** Max 3D centroid separation (mm²) for a "near" match — the general case. */
 const MAX_MATCH_DIST_SQ = 100;
 
-/** Min |normal·normal| for two faces to count as sharing a plane's orientation. */
+/** Min |outNormal·inNormal| for two faces to count as sharing a plane's orientation. */
 const COPLANAR_NORMAL_MIN = 0.985;
 
 /**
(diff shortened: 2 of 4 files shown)


andymai:
@cubic-dev-ai review

PR #1799, opened 2026-07-16
fix: plug arena leaks in the projection path (makeProjectedEdges, drawProjection)

andymai:
Continues the arena-leak sweep into the **assembly / projection / text** families (probed all three). The leaks were in **projection**; assembly and base text were already clean. Oracle now **51 tests**.

## Fixed (measured via `getShapeCount()`)

| Op | Before → after | Root cause |
|---|---|---|
| **makeProjectedEdges** | 12 → 0 | the 6 HLR result compounds from `projectEdges` were never disposed; returned edges were *borrowed* from their (leaked) topology caches. Now manages each compound with `using` and returns independent **clones** the caller owns. |
| **drawProjection** | 22 → 0 | never disposed the projected edges it consumes, and `edgesToDrawing` leaked the throwaway `drawRectangle().sketchOnPlane()` plane wire per call. |

## Already clean (verified, no change)
- `solveAssembly` — pure transform math, allocates no shapes (0/call)
- `textBlueprints`, `drawText`, and `sketchText` for straight/open glyphs (I, C, LT) — 0/call

## Known residual (documented, not fixed)
`sketchText` leaks **1 slot for a glyph that contains a hole** (B, O, …) — an intermediate in the 2D hole-contour wire path, independent of the returned contour wires (disposing every returned wire/face still leaves 1). Straight/open glyphs are clean. Left as a documented follow-up; it needs a font to reproduce so it's not in the arena oracle.

## Verification
- `vitest run --project occt-wasm tests/wasmArenaDisposal.test.ts` → 51/51
- 85 tests pass across projection/draw3d/text/assembly suites
- typecheck / eslint / prettier / pattern-check clean; total-JS 344.6 kB (within 345 kB)

<a href="https://cubic.dev/pr/andymai/brepjs/pull/1799?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/projection/makeProjectedEdges.ts
+++ b/src/projection/makeProjectedEdges.ts
@@ -41,13 +41,22 @@ export function makeProjectedEdges(
     ...getEdgesFromOc(projected.visible.outline),
   ];
 
-  const hidden = withHiddenLines
-    ? [
-        ...getEdgesFromOc(projected.hidden.sharp),
-        ...getEdgesFromOc(projected.hidden.smooth),
-        ...getEdgesFromOc(projected.hidden.outline),
-      ]
-    : [];
+  // projectEdges always allocates all six result compounds; getEdgesFromOc is the
+  // only thing that disposes them (via `using compound`). So extract the hidden
+  // compounds even when unwanted — then dispose the clones — rather than skipping
+  // the calls and leaking the three hidden compounds' arena slots.
+  const hiddenEdges = [
+    ...getEdgesFromOc(projected.hidden.sharp),
+    ...getEdgesFromOc(projected.hidden.smooth),
+    ...getEdgesFromOc(projected.hidden.outline),
+  ];
+  let hidden: Edge[];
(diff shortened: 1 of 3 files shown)


andymai:
Good catch @greptile — fixed in e83b5e38. You're right that `projectEdges` unconditionally allocates all six compounds, so the `withHiddenLines=false` branch was leaking the three hidden ones (getEdgesFromOc's `using compound` is the only disposal). Now the hidden compounds are always extracted (disposing their slots) and the unwanted clones are disposed. Added a second oracle assertion covering `withHiddenLines=false` at 0 slots/call.

PR #1803, opened 2026-07-16
test(sketching): prove sketchText is leak-free; mark .wires accessors disposable

andymai:
Follow-up to the "sketchText hole-glyph leak" documented in #1799. **Investigated it — it's not a real leak.**

## Finding

Stage isolation on `sketchText('O')` (a hole glyph):

| Stage | Leak/call |
|---|---|
| `textBlueprints('O')` | 0 |
| `textBp.sketchOnPlane('O')` + dispose wires | 0 |
| `sketchText('O')` + dispose via sub-sketch wires | **0** |
| `sketchText('O')` + probe reads `.wires` getter | 1 |

The extra slot only appears when the disposal walk **reads `CompoundSketch.wires`** — a **getter** that calls `makeCompound(...)` and allocates a fresh compound on every access — and drops the returned compound. My earlier probe's generic `disposeAllWires` helper touched that getter, so it *measured* a leak that sketchText never had. No production code reads `.wires` transiently.

## Changes (no behavior change)

- **Document the footgun**: `CompoundSketch.wires` (getter) and `Sketches.wires()` (method) now carry `@remarks` that they allocate a **caller-owned** kernel compound on access that must be disposed. A side-effecting getter that allocates WASM memory is the real hazard here.
- **Regression tests** (occt-wasm arena oracle, font-gated): a hole glyph disposed via its contour wires leaks nothing, and the `.wires` compound is disposable back to baseline. Oracle 51 → 53 tests.

## Verification
- `vitest run --project occt-wasm tests/wasmArenaDisposal.test.ts` → 53/53 (2 new pass with a system mono font; skip cleanly without one)
- typecheck / eslint / prettier clean

<a href="https://cubic.dev/pr/andymai/brepjs/pull/1803?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/sketching/compoundSketch.ts
+++ b/src/sketching/compoundSketch.ts
@@ -48,7 +48,15 @@ export default class CompoundSketch implements SketchInterface {
     return this.sketches.slice(1);
   }
 
-  /** Return all wires (outer + holes) combined into a compound shape. */
+  /**
+   * All wires (outer + holes) combined into a compound shape.
+   *
+   * @remarks Allocates a **fresh** compound on every access (via `makeCompound`)
+   * — it is a caller-owned kernel resource, not a cheap accessor. Dispose the
+   * returned compound (`using`/`.delete()` → `Symbol.dispose`) or it leaks an
+   * arena slot on arena kernels. The sub-sketch wires themselves belong to this
+   * CompoundSketch and are freed with it.
+   */
   get wires() {
     return fns.compoundSketchWires(this);
   }
--- a/src/sketching/sketches.ts
+++ b/src/sketching/sketches.ts
@@ -69,15 +69,19 @@ export default class Sketches implements SketchInterface {
   }
 
   /**
-   * All wires combined into a single compound shape.
+   * All contour wires combined into a single compound shape.
    *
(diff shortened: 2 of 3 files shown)


andymai:
Good catch @greptile — you're right, that was a genuine leak, not just a doc gap. Fixed in 71ed6caa.

`Sketches.wires()` now flattens to the borrowed per-contour wires (reading each CompoundSketch's sub-sketch `.wire` directly) instead of calling the allocating `.wires` getter, so no intermediate compound is created for makeCompound to reference-and-drop. Added an oracle regression that disposes everything reachable and asserts 0 — the old unreachable intermediate would have survived it. The only consumers were tests asserting `.toBeDefined()`, and the flat-vs-nested compound is equivalent for "all wires combined".

PR #1829, opened 2026-07-17
refactor(kernel): mark unsupported-operation errors instead of matching messages

### LINKED ISSUE #1825: Replace unsupported-kernel-op error message matching with a marker property
Follow-up from #1824 review (Greptile P2).

`adjacencyFns.isUnsupportedKernelError` classifies "kernel doesn't support this op" by matching three error-message substrings — occt's `makeBrepkitOnlyStubs` (`is only available with the`), occt-wasm's `notImplemented` (`is not yet implemented`), and manifold's `occtOrThrow` (`requires a registered occt kernel`).

This works and fails loudly on a rephrase (tests break rather than silent corruption), but it couples the probe to exact wording across three adapters. The structural fix: have all three throw sites tag the error with a shared marker (e.g. a small `UnsupportedKernelOperationError` or a `isUnsupportedKernelOp` property from a shared kernel-layer helper) and test for that instead of the message. Would also make the 'unsupported operation' concept first-class for any future feature-detected native op.

andymai:
Follow-up to #1824. **Closes #1825.**

## Problem

Feature-detecting a native kernel op vs its JS fallback (in `adjacencyFns.sharedEdges`) was done by matching three adapter error-message substrings. Greptile flagged this as fragile: a reworded message silently breaks the probe, and a genuine geometry error containing one of the phrases would be misclassified as "unsupported" and swallowed into the fallback.

## Change

New `src/kernel/unsupported.ts`:
- **`UnsupportedKernelOperationError`** — carries a `Symbol.for('brepjs.kernel.unsupportedOperation')` marker.
- **`isUnsupportedKernelOperationError(e)`** — checks the marker via `Reflect.get`, **not `instanceof`** (which is unreliable across the multi-entry Vite build's separate dist chunks, where a second copy of the class could be bundled).

The three canonical stub sites now throw it, **messages unchanged** (only the marker is added, so any user or code reading the text is unaffected):
- occt `makeBrepkitOnlyStubs` (`u()`) + inline `meshBoolean`
- occt-wasm `notImplemented`
- manifold `occtOrThrow`

`adjacencyFns` drops the substring regex for the structural guard. "Unsupported operation" is now a first-class, reusable signal for any future feature-detected native op.

## Verification

- Full occt-wasm suite: **4726 pass / 0 fail** (+2 new)
- Adjacency **12/12** on occt-wasm/occt/brepkit — native path on occt-wasm/brepkit, marker-driven JS fallback on occt
- New guard tests assert the marked error is recognized **and** that a plain `Error` merely *reading* like the old sentinels is **not** misclassified (the exact fragility this removes)
- typecheck / lint / check:patterns / check:boundaries / format / knip: clean

<a href="https://cubic.dev/pr/andymai/brepjs/pull/1829?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/kernel/manifold/geometryOps.ts
+++ b/src/kernel/manifold/geometryOps.ts
@@ -15,6 +15,7 @@
 import type { KernelCurveOps } from '@/kernel/interfaces/curveOps.js';
 import type { KernelSurfaceOps } from '@/kernel/interfaces/surfaceOps.js';
 import type { KernelAdapter } from '@/kernel/interfaces/index.js';
+import { UnsupportedKernelOperationError } from '@/kernel/unsupported.js';
 import type { KernelShape } from '@/kernel/types.js';
 import type { ManifoldModule } from './helpers.js';
 import { asManifoldShape, brepCache, occtOrThrow, resolveOcct, unwrap } from './meshHandle.js';
@@ -64,7 +65,9 @@ function viaOcct<T>(
   if (witness && witness.__manifoldSub && witness.occt) {
     const occt = resolveOcct();
     if (!occt) {
-      throw new Error('manifold: sub-shape geometry query requires a registered occt kernel');
+      throw new UnsupportedKernelOperationError(
+        'manifold: sub-shape geometry query requires a registered occt kernel'
+      );
     }
     return query(witness.occt, occt);
   }
@@ -74,7 +77,7 @@ function viaOcct<T>(
   }
   const occt = resolveOcct();
   if (!occt) {
-    throw new Error(
+    throw new UnsupportedKernelOperationError(
       'manifold: exact geometry query unsupported on manifold kernel; no B-rep kernel registered'
     );
   }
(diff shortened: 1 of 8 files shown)


andymai:
Good catch on the two manifold paths — addressed in the follow-up commit. `geometryOps.viaOcct` (both no-occt guards) and `measureOps.surfaceCurvature` now throw `UnsupportedKernelOperationError` for the no-occt-registered condition, so the whole 'kernel can't do this op' surface is uniformly detectable. I intentionally left two throws as plain `Error`: the bad-shape-handle guard (invalid input, not an unsupported op) and the non-replayable-shape guard (data-dependent — marking it would let a future feature-probe wrongly memoize the op as unsupported for *all* shapes, the same hazard the sharedEdges memo avoids).

PR #1907, opened 2026-07-30
fix(2d): free 2D handles stranded by custom corners

### LINKED ISSUE #1906: Curve2D.boundingBox caches a BoundingBox2d that Curve2D disposal never frees (WASM handle leak in every intersectCurves caller)
## Summary

`Curve2D.boundingBox` lazily creates **and caches** a `BoundingBox2d` (a `Bnd_Box2d` kernel handle), but `Curve2D` disposal never frees that cache. Any code that reads `curve.boundingBox` therefore leaks a kernel handle to the `FinalizationRegistry` — it's reclaimed only on GC, never by explicit disposal.

The most-hit path is `intersectCurves()`, whose AABB fast-reject reads **both** operands' bounding boxes:

```js
var intersectCurves = (first, second, precision = 1e-9) => {
  if (first.boundingBox.isOut(second.boundingBox)) return ok({ intersections: [], ... });
  ...
};
```

Each call leaks **2** `BoundingBox2d` handles. Since `filletCurves → removeCorner → intersectCurves` runs once per rounded corner, `DrawingPen.customCorner()` leaks exactly 2 handles per corner — and the same applies to `chamferCurves`, `dogboneFilletCurves`, and every drawing boolean that routes through `intersectCurves`.

## The cache

```js
// Curve2D
get boundingBox() {
  if (this._boundingBox) return this._boundingBox;
  const kernel = getKernel();
  const boundBox = kernel.createBoundingBox2d();
  kernel.addCurveToBBox2d(boundBox, this.wrapped, 1e-6);
  this._boundingBox = new BoundingBox2d(boundBox); // registerForCleanup → GC only
  return this._boundingBox;
}
```

`new BoundingBox2d(...)` calls `registerForCleanup(this, this._wrapped)`, so the handle is freed only when the JS wrapper is GC'd. `Curve2D`'s own `delete()`/`[Symbol.dispose]` frees `this.wrapped` but not `this._boundingBox`, so even explicitly disposing the curve leaves the cached box to leak.

## Reproduction

Isolate with a scope (which disposes everything the pen registers) and count `gcCollected`, so only the unregistered leak shows:

```js
async function leakPerCall(build, n = 20) {
  withScope(() => build());
  await drainGc(); // gc(); await microtask; gc(); await microtask
  const start = getDisposalStats().gcCollected;
  for (let i = 0; i < n; i++) withScope(() => build());
  await drainGc();
  return (getDisposalStats().gcCollected - start) / n;
}

// run under NODE_OPTIONS=--expose-gc
const sharp    = await leakPerCall(() => draw([0,0]).lineTo([10,0]).lineTo([10,10]).lineTo([0,10]).close());       // 0/call
const filleted = await leakPerCall(() => draw([0,0]).lineTo([10,0]).customCorner(2).lineTo([10,10]).lineTo([0,10]).close()); // 2/call
```

Wrapping `registry.register` to tag each handle with `new Error().stack` and logging it in the `FinalizationRegistry` callback gives the exact site:

```
new BoundingBox2d ← Curve2D.get boundingBox ← intersectCurves ← removeCorner ← filletCurves ← DrawingPen._nextCorner
```

## Impact

Unbounded kernel-handle growth for any long-running session that repeatedly builds filleted/chamfered drawings or drawing booleans. Surfaced downstream (gridfinity-layout-tool) as `compartments delta=280` in a disposal-regression harness and gradual generation-worker memory growth.

## Suggested fixes (either, ideally the first)

1. **Make `Curve2D` disposal free its cached `_boundingBox`.** In `delete()`/`[Symbol.dispose]`, if `this._boundingBox` exists, dispose it and null the cache. This fixes the leak for *all* `boundingBox` readers, provided the curve is eventually disposed.

2. **Have `intersectCurves` dispose the boxes it triggers.** Capture both, run `isOut`, clear the operands' `_boundingBox` cache, and delete them. Fixes the hottest path regardless of curve lifetime. (This is the shape of the downstream `pnpm patch` workaround currently in use.)

Both keep geometry identical — it's purely a disposal change.

## Downstream workaround

gridfinity-layout-tool currently ships a `pnpm patch` implementing fix #2 in the two `dist/blueprintSketcher-*` bundles, to be retired once this is fixed upstream.

andymai:
Fixes #1906.

## What leaked

`Curve2D.boundingBox` lazily creates **and caches** a `BoundingBox2d`, but `Curve2D` disposal never freed the cache, so every `intersectCurves` caller stranded two boxes in the `FinalizationRegistry`.

Fixing only the cache was not enough. Measuring with a `gcCollected` probe showed a filleted corner still leaking 2 handles per call, because the custom-corner path never disposed the curves whose boxes those were:

- `removeCorner` / `dogboneFilletCurves` build offset curves (and, for dogbone, a locator circle) purely to find the corner centre, then drop them.
- `intersectCurves` returns `commonSegments` as `Curve2D` handles that these callers discard.
- `splitAt` returns both halves; only one is kept per side.
- `BaseSketcher2d` pops the two curves a corner function consumes and pushes the replacements without deleting the originals.

## Changes

- `Curve2D.delete()` disposes and clears its cached bounding box.
- `removeCorner` / `dogboneFilletCurves` dispose their offsets, locator circle, unused split pieces, and common segments on every exit path (`try`/`finally`, so the early-bail branches are covered too).
- `BaseSketcher2d` deletes the curves a corner function consumed, skipping any the function returned untouched (corner functions return their inputs verbatim when they bail on collinear curves or a failed offset).
- `--expose-gc` added to the vitest `execArgv` so leak probes can force collection.

I did not take the issue's suggested fix #2 (clearing the operands' cache inside `intersectCurves`). `boolean2D` intersects each curve against every other, so per-call invalidation would turn O(n) box builds into O(n²). Ownership at the producer is both cheaper and more general.

## Verification

`gcCollected` delta per call, 20 iterations, occt-wasm:

| drawing | before | after |
| --- | --- | --- |
| sharp corners (baseline) | 0 | 0 |
| `customCorner(2)` fillet | 2 | 0 |
| `customCorner(2, 'chamfer')` | 2 | 0 |
| `closeWithCustomCorner(2, 'dogbone')` | 2 | 0 |

- `tests/customCornerHandleLeaks.test.ts` (new): the four probes above, plus geometry assertions that fillet/chamfer/dogbone still return three connected curves and that the collinear bail path returns usable curves. Geometry block is skipped on manifold, which has no 2D offset.
- `tests/curve2dFns.test.ts`: deterministic (no-GC) assertions that a box read from a curve is disposed with it and that `delete()` stays idempotent.
- Full occt-wasm suite green (3941 passed). The 2D suites were also run across all four kernel projects: the only failures are the 86 pre-existing manifold ones, identical on `main`.

## Behavior note

`Curve2D.boundingBox` is now **borrowed**: it is disposed with the curve. Documented on the getter, on `curve2dBoundingBox`, and in `src/2d/README.md`. Callers needing an independent box should clone the curve first.

The sketcher now deletes curves it hands to a *user-supplied* corner function (`customCorner` accepts a callback). A callback that retains its input curves past the call would see them disposed; returning them in the result array opts out.

## Downstream

`gridfinity-layout-tool`'s `pnpm patch` on `dist/blueprintSketcher-*` can be retired once this ships.

### DIFF
--- a/src/2d/README.md
+++ b/src/2d/README.md
@@ -78,3 +78,4 @@ graph TB
 5. **Serialization**: `Curve2D.serialize()` enables persistence; curves can be saved and restored
 6. **Kernel-abstracted**: All 2D curve operations (evaluation, type queries, circle/ellipse data, Bezier poles, bounding boxes) go through `getKernel()` / `getKernel2D()` methods. No direct kernel API calls in this module.
 7. **Disposable**: `Blueprint`, `Curve2D`, and `BoundingBox2d` implement `Symbol.dispose`, enabling `using` keyword for deterministic WASM cleanup: `using bp = new BlueprintSketcher().movePointerTo([0,0]).lineTo([10,0]).close();`
+8. **`boundingBox` is borrowed**: `Curve2D.boundingBox` caches its `BoundingBox2d` and disposes it with the curve, so a box read from a curve must not outlive it. Clone the curve first if you need an independent box.
--- a/src/2d/blueprints/baseSketcher2d.ts
+++ b/src/2d/blueprints/baseSketcher2d.ts
@@ -27,6 +27,17 @@ const cornerModeFns = {
   fillet: filletCurves,
 } as const;
 
+/**
+ * Delete the curves a corner function consumed. Corner functions return their
+ * inputs untouched when they bail (collinear curves, failed offset), so only
+ * the ones absent from the result are ours to free.
+ */
+function disposeConsumedCurves(consumed: Curve2D[], produced: Curve2D[]): void {
+  consumed.forEach((curve) => {
+    if (!produced.includes(curve)) curve.delete();
+  });
+}
+
 function buildCornerFunction(
   radius: number | ((first: Curve2D, second: Curve2D) => Curve2D[]),
   mode: 'chamfer' | 'fillet' | 'dogbone'
@@ -131,7 +142,9 @@ export class BaseSketcher2d {
     if (!previousCurve)
       bug('Sketcher2d.saveCurve', 'No previous curve available for custom corner');
 
-    this.pendingCurves.push(...this._nextCorner(previousCurve, curve));
+    const cornered = this._nextCorner(previousCurve, curve);
+    disposeConsumedCurves([previousCurve, curve], cornered);
+    this.pendingCurves.push(...cornered);
     this._nextCorner = null;
   }
 
@@ -435,7 +448,9 @@ export class BaseSketcher2d {
     if (!previousCurve || !curve)
       bug('Sketcher2d._customCornerLastWithFirst', 'Not enough curves to close and fillet');
 
-    this.pendingCurves.push(...buildCornerFunction(radius, mode)(previousCurve, curve));
+    const cornered = buildCornerFunction(radius, mode)(previousCurve, curve);
+    disposeConsumedCurves([previousCurve, curve], cornered);
+    this.pendingCurves.push(...cornered);
   }
 
   protected _closeSketch(): void {
--- a/src/2d/lib/curve2D.ts
+++ b/src/2d/lib/curve2D.ts
@@ -58,6 +58,10 @@ export class Curve2D {
   delete(): void {
     if (!this._deleted) {
       this._deleted = true;
+      if (this._boundingBox) {
+        this._boundingBox.delete();
+        this._boundingBox = null;
+      }
       if (typeof this._wrapped.delete === 'function') {
         unregisterFromCleanup(this._wrapped);
         this._wrapped.delete();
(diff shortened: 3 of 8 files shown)


andymai:
Both cases are real, fixed in bfe3c759.

- **Offsets**: `make2dOffset` for both curves now runs inside the guarded region in `removeCorner` and `dogboneFilletCurves`, so a throw on the second one still frees the first.
- **Splits**: the second split runs through `splitOrDiscard`, which disposes the first curve's pieces before rethrowing.
- Also covered the same class of gap in `filletCurves` / `chamferCurves`: `make2dArcFromCenter` / `make2dSegmentCurve` can throw after the trims exist, so the trims are disposed before the error propagates (it still propagates, no behavior change).

Replacing `wasmIndex` with a checked read to do this surfaced a live bug: a radius that consumes a whole segment splits it at its own endpoint, so `splitAt` returns the curve itself and `secondSplit[1]` is `undefined`. `wasmIndex` cast that away and `filletCurves` threw `TypeError: Cannot read properties of undefined (reading 'firstPoint')`. On a 10x10 rectangle, `customCorner(10)` reproduces it on `main`. It now bails to the untouched curves, the same contract the collinear case already has, with a regression test.

PR #1912, opened 2026-07-30
fix(2d): make Drawing and blueprint containers disposable

### LINKED ISSUE #1910: Blueprints/CompoundBlueprint/Drawing cache a BoundingBox2d with no disposal path
## Summary

`Blueprints` and `CompoundBlueprint` cache a `BoundingBox2d` exactly the way `Blueprint` does, but unlike `Blueprint` they have **no `delete()` / `[Symbol.dispose]` at all**, so the cached box is never disposed. `Drawing`, the public result of `draw()`, has no disposal either, so `Blueprint.delete()` is effectively unreachable from the public API and the whole 2D tree is left to GC.

Found while fixing #1906.

## The gap

```ts
// blueprints.ts:29 / compoundBlueprint.ts:34
protected _boundingBox: BoundingBox2d | null;

get boundingBox(): BoundingBox2d {
  if (!this._boundingBox) {
    const box = new BoundingBox2d();            // registerForCleanup -> GC only
    this.blueprints.forEach((b) => { box.add(b.boundingBox); });
    this._boundingBox = box;
  }
  return this._boundingBox;
}
```

`Blueprint.delete()` (blueprint.ts:105) does the right thing: deletes its curves, then `this._boundingBox.delete()`. The two container classes never got the same treatment, and neither did `Drawing` (drawing.ts:69), which only holds `innerShape`.

Reading `b.boundingBox` inside the loop also populates every child `Blueprint`'s cache, so a container read strands the children's boxes too.

## Severity

Smaller than #1906's, and worth stating precisely: the 2D layer is backed by the pure-TS `geometry2d` module on **all four kernels** (`occt`, `occt-wasm`, `brepkit`, `manifold` all route `createBoundingBox2d` to `src/kernel/geometry2d.ts`), and `createBBox2d()` returns a plain JS object with no `delete`. So no native WASM memory is held today.

What does leak is a `FinalizationRegistry` registration per box (`BoundingBox2d`'s constructor registers unconditionally, regardless of whether the handle has a real `delete`) plus the retained JS objects, both cleared only on GC. It also blocks deterministic cleanup: there is currently no way for a caller to dispose a `Drawing` at all.

## Measurements

`gcCollected` delta per call, 20 iterations, occt-wasm, `--expose-gc`:

| operation | leak/call |
| --- | --- |
| `drawRectangle(20, 2)` | 0 |
| `drawRectangle(20, 2).boundingBox` | 1 |
| `drawRectangle(20, 2).cut(drawRectangle(2, 10))` | 12 |
| ...`.boundingBox` on the resulting `Blueprints` | 13 |
| `Blueprint.intersects(other)` | 5 |

The `+1` on rows 2 and 4 is this issue. The 12 and the 5 are separate, larger sources (below).

## Proposed fix

1. `Blueprints` and `CompoundBlueprint` get `delete()` / `[Symbol.dispose]` that dispose the cached box and cascade to child blueprints, mirroring `Blueprint.delete()`.
2. `Drawing` gets the same, delegating to `innerShape`, so the public tree is disposable and `using` works on a drawing.
3. Regression coverage in `tests/customCornerHandleLeaks.test.ts`'s probe style.

## Out of scope / follow-up

The dominant sources in the table above are not this bug and need their own work:

- **2D booleans strand ~12 registry entries per call.** `boolean2D` builds intermediate blueprints and curves and drops them; the AABB pre-filter at `boolean2D.ts:110` caches a box on each operand. Fixing this needs an aliasing pass over `segmentAssembly` / `intersectionSegments` to establish who owns which curves.
- **`Blueprint.intersects` (blueprint.ts:438)** never disposes the `result.segments` the kernel returns, on either exit path. `isInside` already gets this right.

andymai:
Fixes #1910.

`Blueprint` disposes its cached `BoundingBox2d`; the containers above it never got the same treatment. `Blueprints` and `CompoundBlueprint` cache a box with **no `delete()` / `[Symbol.dispose]` at all**, and `Drawing` has none either, which left `Blueprint.delete()` unreachable from the public API: there was no way for a caller to release a drawing.

## Changes

- `Blueprints` / `CompoundBlueprint` get `delete()` / `[Symbol.dispose]`: dispose the cached box, cascade to child blueprints.
- `Drawing` gets the same via `innerShape`, so `using drawing = drawRectangle(10, 10)` works.
- `Drawing.boundingBox` caches the box it mints for an *empty* drawing instead of returning a fresh, unreachable one on every read.
- `Blueprint.delete()` clears `_boundingBox` rather than leaving a disposed box reachable through the getter.

## Verification

`gcCollected` delta per call, 20 iterations, occt-wasm:

| operation | before | after |
| --- | --- | --- |
| `drawRectangle(20, 2).boundingBox` then dispose | 1 (no way to dispose) | 0 |

`tests/drawingDisposal.test.ts` covers: the empty-drawing box, the cascade from `Drawing` to blueprint curves, idempotent `delete()`, `using`, both container classes, and the GC probes. Full occt-wasm suite green (3902 passed); 2D, playground-example and docs-example suites re-run clean.

## Scope

This is the `+1` row of the table in #1910. The dominant sources in that area are **not** fixed here and remain open:

- 2D booleans strand ~12 registry entries per call from intermediates `boolean2D` drops. Fixing that needs an aliasing pass over `segmentAssembly` / `intersectionSegments` to establish curve ownership.
- `Blueprint.intersects` never disposes the `result.segments` the kernel returns.

## Note on severity

Worth stating plainly: the 2D layer routes through the pure-TS `src/kernel/geometry2d.ts` on all four kernels, and `createBBox2d()` returns a plain JS object with no `delete`. So nothing here was holding native WASM memory. What leaked was a `FinalizationRegistry` registration per box plus retained JS objects, cleared only on GC, and the absence of any deterministic cleanup path for a `Drawing`.

## Behavior note

Disposal cascades to children. Containers built from another container's children (`new Blueprints(other.blueprints.slice(1))`, as `boolean2D` does internally) share those objects, so deleting one invalidates the other. Nothing in the library calls `delete()` on a container, so this is a new capability rather than a changed code path, but it is documented in `src/2d/README.md` and `src/sketching/README.md`.


andymai:
Fair point, done in 74dee2ff. `DrawingInterface` now declares `delete(): void` and `[Symbol.dispose](): void`, with a test that disposes all three implementors through the interface type.

Two notes on the change:

- **Additive for consumers, type-level only for implementors.** All three implementors already satisfied it, so `typecheck` was clean without touching them. It could break an external caller who implements `DrawingInterface` themselves, at compile time only. Nothing in the repo does that outside the three blueprint classes, so I did not mark the commit breaking; happy to re-mark it if the maintainer wants the major bump.
- `Drawing` deliberately does **not** implement `DrawingInterface` (it never did) and is unaffected.

PR #1917, opened 2026-08-01
fix(deps): clear all remaining OSV findings to get main green

andymai:
## Summary

Clears every remaining OSV finding so the blocking `OSV Scan` job on `main` goes green. Verified locally with `osv-scanner` 2.4.0 against the repo's own `osv-scanner.toml`: **`No issues found`**.

Six of the seven are real fixes. One is an explicit, documented exception, because the "fix" available for it would have been cosmetic.

## Real fixes

**sharp and @hono/node-server were held by a parent's range**, so bumping them directly would have needed an override violating that range (the pattern removed in #1916). Both parents had already shipped a fix:

```
sharp:  brepjs-manifold -> manifold-3d -> @gltf-transform/functions
                        -> ndarray-pixels 5.0.1 -> sharp 0.34.5   (range ^0.34.0)
hono:   brepjs-cad -> @modelcontextprotocol/sdk 1.29.0 -> @hono/node-server 1.19.14  (range ^1.19.9)
```

- `ndarray-pixels@5.2.0` requires `sharp ^0.35.0` and satisfies `@gltf-transform/functions`' `^5.0.1`, so it slots in and pulls `sharp@0.35.3`. No `package.json` change: pure lockfile inertia.
- `@modelcontextprotocol/sdk@1.30.0` accepts `@hono/node-server ^1.19.9 || ^2.0.5`, so bumping brepjs-cad's pin reaches `2.0.12`. `@hono/node-server@2` peers on `hono ^4`, satisfied by the installed 4.12.31.

**postcss** 8.5.17 → 8.5.25 and **body-parser** 2.2.2 → 2.3.0 were also plain inertia: `vite` already allowed `^8.5.17` and `express` already allowed `^2.2.1`. The nested `@vue/compiler-sfc` postcss dedupes away now that the hoisted copy satisfies `^8.5.19`.

Clears GHSA-f88m-g3jw-g9cj (7.0 High), GHSA-r28c-9q8g-f849 (7.5 High), GHSA-frvp-7c67-39w9 (5.9), GHSA-v422-hmwv-36x6, plus the transitive `ndarray-pixels` and `@modelcontextprotocol/sdk` alerts. Every High and Moderate is gone via a real bump.

## The dompurify exception, and why not an override

`monaco-editor@0.56.0` is the latest release and pins `dompurify: 3.4.8` exactly. A one-line override would have raised it and turned this table green.

It would also have been meaningless. **That copy is never imported.** monaco vendors its own DOMPurify at `esm/vs/base/browser/dompurify/dompurify.js` (license header: `DOMPurify 3.4.8`), and `domSanitize.js` imports `./dompurify/dompurify.js`. There is no bare `dompurify` import anywhere in monaco's `esm/` or `min/` output. Overriding `node_modules` would bump a package nothing loads, while the vendored copy actually bundled into the playground stayed at 3.4.8 — a green scan over unchanged exposure.

So it is recorded in `osv-scanner.toml` instead, following the existing convention (id + `ignoreUntil` + reason). All three advisories require a DOMPurify API monaco never calls:

| Advisory | Requires | In monaco's `esm/` |
| --- | --- | --- |
| GHSA-cmwh-pvxp-8882 (5.1) | `setConfig()` | only DOMPurify's own definition; the other `setConfig` hits are unrelated same-named methods in the TS language service and monaco's config layer |
| GHSA-c2j3-45gr-mqc4 (2.1) | `CUSTOM_ELEMENT_HANDLING` | only inside the vendored `dompurify.js`; never configured |
| GHSA-vxr8-fq34-vvx9 (2.1) | `clearConfig()` | only inside the vendored `dompurify.js`; never called |

monaco passes config per-call to `purify.sanitize(...)` and calls `purify.removeAllHooks()` in a `finally`, so the cross-call config persistence all three rely on never happens. Verified against the vendored source, not just `package.json`. Entries expire **2026-11-01**; bumping monaco is the only real fix and upstream has not shipped one.

## On the lockfile

Generated with **npm 11.6.1**, matching the repo's `"packageManager"`. A newer npm additionally flips `"peer": true` on ~40 unrelated packages.

Most of the 643 lines are `sharp` 0.35's wider platform matrix (`@img/sharp-*` gained freebsd-wasm32, linux-ppc64/riscv64/s390x, linuxmusl-arm64, webcontainers-wasm32, win32-arm64/ia32, darwin). The lockfile previously carried only the linux-x64 pair. Excluding `@img/*`, the complete set of version changes is nine:

```
@hono/node-server      1.19.14 -> 2.0.12
@modelcontextprotocol/sdk 1.29.0 -> 1.30.0
body-parser            2.2.2   -> 2.3.0     (+ nested content-type 2.0.0)
ndarray-pixels         5.0.1   -> 5.2.0
postcss                8.5.17  -> 8.5.25
semver                 7.7.4   -> 7.8.5     (required by sharp 0.35.3: semver ^7.8.5)
sharp                  0.34.5  -> 0.35.3
@vue/compiler-sfc/postcss  8.5.24 -> removed (dedupes to hoisted)
```

One trap worth flagging: `npm update postcss` re-resolved the **vitepress subtree** as a side effect and downgraded its pinned `vite` 6.4.3 → 5.4.21 and `esbuild` 0.25.10 → 0.21.5, which surfaced four *new* advisories. That subtree is restored to `main`'s exact state; the docs build is verified below.

## Verification

- [x] `osv-scanner scan --lockfile=package-lock.json` → **No issues found**
- [x] `npm ci` (11.6.1) clean
- [x] `sharp@0.35.3` native binary loads and encodes a real PNG (libvips 8.18.3)
- [x] `ndarray-pixels@5.2.0` + new sharp round-trips decode → encode
- [x] `npm run build` (library) clean
- [x] `npm run build --workspace=apps/playground` clean — monaco bundles fine
- [x] `npm run docs:build` clean — confirms the restored vitepress subtree works
- [x] `brepjs-cad` typecheck + tests: 31 files, 195 tests passed against SDK 1.30.0
- [x] `tests/manifoldCopyShape.test.ts` passed
- [x] root `typecheck`, `lint`, `check:boundaries`, `format:check` clean

## Left alone

`osv-scanner.toml`'s existing `GHSA-gv7w-rqvm-qjhr` (esbuild) entry is now reported as an **unused ignore** — the esbuild overrides appear to have resolved it. Its `ignoreUntil` is 2026-09-15, so it will force a re-look shortly; removing it is out of scope here.

Two lockfile fields are deliberately kept at `main`'s values rather than npm's output: `node_modules/brepjs` `version` and `packages/brepjs-cad` → `brepjs` range. `packages/brepjs-cad/package.json` declares `">=18.117.1"` while release-please bumps that range in the lockfile only, so every `npm install` reconciles it back and every release flips it forward. Worth fixing at the release-please config, separately.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -2,7 +2,7 @@
   ".": "18.119.5",
   "packages/brepjs-opencascade": "0.16.0",
   "packages/brepjs-voxel-wasm": "0.7.0",
-  "packages/brepjs-cad": "0.107.1",
+  "packages/brepjs-cad": "0.108.0",
   "packages/brepjs-bim": "0.3.1",
   "packages/brepjs-sheetmetal": "0.3.0"
 }
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,8 +1,8 @@
 {
-  ".": "18.119.4",
+  ".": "18.119.5",
   "packages/brepjs-opencascade": "0.16.0",
   "packages/brepjs-voxel-wasm": "0.7.0",
-  "packages/brepjs-cad": "0.107.0",
+  "packages/brepjs-cad": "0.107.1",
   "packages/brepjs-bim": "0.3.1",
   "packages/brepjs-sheetmetal": "0.3.0"
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [18.119.5](https://github.com/andymai/brepjs/compare/brepjs-v18.119.4...brepjs-v18.119.5) (2026-08-01)
+
+
+### Bug Fixes
+
+* **deps:** drop redundant brace-expansion override ([#1916](https://github.com/andymai/brepjs/issues/1916)) ([b63686b](https://github.com/andymai/brepjs/commit/b63686bfff611fe753ed63c7ace3586199270499))
+
 ## [18.119.4](https://github.com/andymai/brepjs/compare/brepjs-v18.119.3...brepjs-v18.119.4) (2026-07-31)
 
 
--- a/osv-scanner.toml
+++ b/osv-scanner.toml
@@ -14,23 +14,36 @@ ignoreUntil = 2026-09-15
 reason = "esbuild dev-only via vitepress; Deno/NPM_CONFIG_REGISTRY vector not reachable in Node CI; fix breaks docs build"
 
 # --- dompurify via monaco-editor (3 advisories) ---------------------------
-# monaco-editor 0.56.0 (latest; no newer release) declares `dompurify: 3.4.8`
-# as an exact pin, which is what the scanner matches on. That copy is never
-# loaded: monaco vendors its own DOMPurify at
-# esm/vs/base/browser/dompurify/dompurify.js (license header: DOMPurify 3.4.8)
-# and domSanitize.js imports `./dompurify/dompurify.js`. There is no bare
-# `dompurify` import anywhere in monaco's esm/ or min/ output.
+# monaco-editor 0.56.0 declares `dompurify: 3.4.8` as an exact pin, which is
+# what the scanner matches on. That copy is never executed, for two reasons.
 #
-# So an npm override would raise the unused node_modules copy and turn this
-# table green while the vendored copy actually shipped in the playground
-# bundle stayed at 3.4.8. That is worse than an explicit exception: it hides
-# the exposure instead of recording it. Bumping monaco is the only real fix
-# and upstream has not shipped one.
+# 1. monaco never imports the package. It vendors its own DOMPurify at
+#    esm/vs/base/browser/dompurify/dompurify.js and domSanitize.js imports
+#    `./dompurify/dompurify.js`. There is no bare `dompurify` import anywhere
+#    in monaco's esm/ or min/ output.
+# 2. The playground never bundles monaco at all. It uses @monaco-editor/react,
+#    and with no loader.config() override the default applies: monaco is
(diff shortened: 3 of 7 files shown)


andymai:
### Correction to the dompurify rationale (012f5066)

While double-checking the exception I'd written, I found one of its claims was wrong, so the `osv-scanner.toml` comment is corrected.

**What I said:** the vendored DOMPurify "actually shipped in the playground bundle stayed at 3.4.8."

**What's true:** `apps/playground` uses `@monaco-editor/react` with no `loader.config()` override anywhere, so the default applies and monaco is fetched at runtime from `cdn.jsdelivr.net`. Monaco is never bundled — there is no DOMPurify in any `apps/playground/dist` chunk, and the emitted `monaco-*.js` is 21 kB (the React wrapper and setup only). `node_modules/monaco-editor` is a **build-time, types-only** dependency here.

The conclusion is unchanged and in fact stronger: an override would raise a copy that is neither imported **nor shipped**.

### A separate finding this surfaced

Because monaco is CDN-loaded, the editor actually executing in the playground is whatever `@monaco-editor/loader` defaults to — **0.55.1** — not the **0.56.0** pinned in `apps/playground/package.json`. Those are two different editors, and the lockfile pin has no effect on what users run.

`monaco-editor@0.55.1` vendors **DOMPurify 3.2.7**, i.e. older than the 3.4.8 flagged above. OSV cannot see this: it only reads `package-lock.json`, so the version that actually runs is outside its view entirely.

Not addressed in this PR — it is a monaco-version and supply-chain question, not something an override or an ignore entry can fix. Worth its own issue. The two obvious options are pinning the loader to the same version as the package (`loader.config({ paths: { vs: ... } })`) or bundling monaco locally so the lockfile pin becomes real.

PR #1925, opened 2026-08-02
fix(brepkit): transform compounds in applyMatrix

andymai:
## What

Rotating or translating a compound threw `brepkit: applyMatrix does not support 'compound' shapes`. `applyMatrix` handled solid/face/wire/edge and fell through to a throw for compounds.

## Why it matters

This accounted for **18 of the 21 failures** across two whole gridfinity scenario families. Both builders hit it the same way — `floorPatternBuilder` does `rotate(panel, -90, …)` then `translate(laid, …)`, and `dividerPatternBuilder` the same, where `panel` is a compound of cutter parts. The floor builder explicitly reuses the divider builder's panel factory, which is why one gap took out both families.

The throw happens *before any geometry runs*, so no boolean was ever reached — the floor file was failing 9 of 11 tests in 5.4 seconds.

## Fix

Handle both compound flavours, because `makeCompound` chooses between them:

- **all-solid** → a real arena compound: `getCompoundSolids` → `copySolid` + `transformSolid` → `makeCompound`
- **anything else** → a JS-side `syntheticCompounds` entry: transform each child and re-register under a fresh synthetic id

Covering only the arena case would silently drop wire and face children.

## Measured, tool-side

With the built `dist/` overlaid into the tool's resolved `brepjs`:

| family | before | after |
|---|---|---|
| floor patterns | 9 failed / 2 passed | **1 failed / 10 passed** |
| divider patterns | 12 failed / 3 passed | **3 failed / 12 passed** |

**21 failures → 4.** Almost none was hiding a second defect: 17 pass outright once they reach the kernel and export watertight. The floor file's runtime goes 5.4 s → 71 s for exactly that reason — those tests previously threw before doing any work.

The 4 remaining are genuine geometry defects, now unblocked and visible.

## Verification

Four tests in `tests/brepkitExtended.test.ts` (which runs under the `brepkit` project), covering both compound flavours, non-mutation of the source, and that every member survives — all differentially verified to fail with the exact production error without the fix.

Suite A/B on `--project brepkit`: **identical failure set**, treatment +4 passing (3595 vs 3591). The 86 pre-existing file-level failures are untouched.

Note: `tests/brepkit-adapter.test.ts` looks like the natural home for these but is excluded from every vitest project, so tests added there never run.

### DIFF
--- a/src/kernel/brepkit/internalOps.ts
+++ b/src/kernel/brepkit/internalOps.ts
@@ -16,6 +16,10 @@ import {
   faceHandle,
   edgeHandle,
   wireHandle,
+  compoundHandle,
+  syntheticCompounds,
+  nextSyntheticId,
+  toArray,
   unwrap,
   DEFAULT_DEFLECTION,
 } from './helpers.js';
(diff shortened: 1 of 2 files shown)


andymai:
Both review points were correct and are fixed in 24a56aa4.

**Vertex children threw.** The recursive compound path sends every child back through `applyMatrix`, and the public `compound()` accepts vertices, so a synthetic compound holding one hit the `default` arm. Added a `vertex` case — a vertex has no surface to offset, just a point to move (`getVertexPosition` → transform → `makeVertex`).

**The synthetic path was untested.** Correct: all four fixtures were all-solid, so every one took the arena path. Added two synthetic-path tests, one with a face child and one with a vertex child.

Differentially verified: with the `vertex` arm disabled the new test fails with exactly `brepkit: applyMatrix does not support 'vertex' shapes`.

Suite: 3597 passed (was 3595, i.e. +2 for the new tests), same single pre-existing failure, 86 pre-existing file-level failures unchanged.

PR #1982, opened 2026-08-09
fix(manifold): transform profile shapes by baking the matrix into their points

andymai:
## Problem

Profile shapes (wires and faces built from points) are backed by an inert placeholder rather than a Manifold solid, so `applyMatrix` crashed with `solid.transform is not a function`. Any transform of a profile was a hard failure on the manifold kernel.

## Fix

Bake the matrix into the recorded points and emit a node of the same kind. Replay rebuilds profile nodes from those same points, so the transform applies exactly once.

The detection matters more than the math. The first version matched the node's `params` against known profile point keys (`pts`, `ring`, `outline`). That over-matched: solid-producing nodes also carry point arrays, so real solids were diverted into the profile path and never actually transformed. Net effect was 632 to 639 failures, and the totals hid it: 3 genuine fixes against 10 new breakages.

Probing the backing object for `transform()` instead is exactly the condition that produces the crash. A real solid always has the method, so the guard cannot misfire in the silent direction. The asymmetry is deliberate: wrongly taking the profile path drops a transform quietly, wrongly taking the solid path throws.

Point handling now goes through an `isVec3` type guard, which also verifies elements are numbers and removes two double-casts.

## Verification

| run | manifold failures |
| --- | --- |
| baseline | 632 |
| params-shape predicate | 639 (3 fixed, 10 broken) |
| backing-object predicate | **629 (3 fixed, 0 broken)** |

Fixed: `compoundOps` boss/pocket, `sketcher3d closeWithMirror`.

The type-guard refactor was re-checked against the five affected files: 16 failures before and after, so it is behavior-preserving.

Default occt-wasm gate: 4222 passed, 0 failed.

### DIFF
--- a/src/kernel/manifold/measureOps.ts
+++ b/src/kernel/manifold/measureOps.ts
@@ -289,8 +289,27 @@ export function makeMeasureOps(_module: ManifoldModule): KernelMeasureOps {
       if (e && e.__nativeEdge && typeof e.length === 'number') return e.length;
       // Standalone profile edges carry an exact analytic descriptor.
       const ms = asManifoldShape(shape);
-      const node = ms?.node as { op?: string; params?: { curve?: CurveDesc } } | undefined;
+      const node = ms?.node as
+        | { op?: string; params?: { curve?: CurveDesc; pts?: ReadonlyArray<readonly number[]> } }
+        | undefined;
       if (node?.op === 'profileEdge' && node.params?.curve) return descLength(node.params.curve);
+      // No descriptor (a transform the descriptor's form could not represent
+      // dropped it) — measure the sampled polyline, which is the geometry that
+      // actually travelled with the shape.
+      const pts = node?.op === 'profileEdge' ? node.params?.pts : undefined;
+      if (pts && pts.length > 1) {
+        let total = 0;
+        for (let i = 1; i < pts.length; i++) {
+          const a = pts[i - 1] ?? [];
+          const b = pts[i] ?? [];
+          total += Math.hypot(
+            (b[0] ?? 0) - (a[0] ?? 0),
+            (b[1] ?? 0) - (a[1] ?? 0),
+            (b[2] ?? 0) - (a[2] ?? 0)
+          );
+        }
+        return total;
+      }
       return notImplemented('length');
     },
     centerOfMass: (shape) => centerOfMass(shape),
--- a/src/kernel/manifold/replay.ts
+++ b/src/kernel/manifold/replay.ts
@@ -313,6 +313,22 @@ const HANDLERS: Readonly<Record<string, ReplayHandler>> = {
   profileEdge: (t, p) => {
     const pts = (p['pts'] as Array<readonly [number, number, number]> | undefined) ?? [];
     const a = pts[0] ?? [0, 0, 0];
+    // Interpolate through every sample when there are interior points. Taking
+    // only the endpoints turns a sampled curve into its chord, and a closed one
+    // (first === last) into a degenerate edge — so a profile whose analytic
+    // descriptor is absent would replay as the wrong shape rather than as a
+    // lower-precision version of the right one.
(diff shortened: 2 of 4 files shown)


andymai:
Addressed the P1 in 0ded6f22.

A profile edge carries an exact analytic curve descriptor that `geometryOps.ts:102` and `measureOps.ts:293` prefer over the sampled points, so baking the matrix into the points alone left curve-point queries answering in the source frame and length ignoring scale, exactly as described.

The variants do not transform alike:

- **line, bezier** are defined purely by points and are affine-invariant, so they transform under any matrix, including shears.
- **conic, helix** store `rx`/`ry`/`radius`/`pitch` as scalars against unit axes. That form only survives a similarity, since shear turns a circle into an ellipse whose axes those fields cannot express. `similarityScale` checks the linear part's columns for equal length and mutual orthogonality and returns the factor only when the matrix qualifies.

When the descriptor cannot be expressed in its own form it is dropped rather than retained, so queries fall back to the sample points that the same function already transformed. A dropped descriptor costs precision; a stale one answers confidently in the old space.

Regression coverage in `tests/manifoldProfileTransform.test.ts`: circle length under scale (the reported case), line length under a non-uniform stretch (point-defined curves still transform exactly rather than being conservatively dropped), and shear on a circle (descriptor dropped rather than reporting the original circumference). Three of the four fail without the fix.

Manifold suite: 629 to 620 failures, no regressions. Beyond the reported case this also fixes six `loft` tests and two `compoundOpsFns` tests that were consuming stale descriptors from transformed profiles.

andymai:
Good catch, and it invalidates the justification I gave for dropping.

I argued that dropping a non-representable descriptor was safe because queries would fall back to the sampled points. That was wrong. `replay.ts:313` rebuilt a `profileEdge` from `pts[0]` and `pts[last]` as a line edge, discarding every interior point — so a sheared circle replayed as its chord, and since a closed curve has `first === last`, a degenerate one. `measureOps.length` threw outright with no descriptor. That trades a wrong-scale answer for wrong-shape geometry, which is worse.

The premise was fine; the fallback did not exist. Built in d6927587:

- **replay** interpolates a spline through all samples when there are interior points, periodic when the curve closes, instead of taking the chord.
- **length** measures the sampled polyline when no descriptor survives, instead of throwing.

The shear test previously wrapped `length` in try/catch, which wrote the test around the weakness. It now asserts the real value: a radius-5 circle stretched 2x in x is an ellipse with semi-axes 10 and 5, perimeter ~48.4 by Ramanujan. It passes.

Manifold suite: 620 failures before and after, no regressions. The replay change touches every `profileEdge` reconstruction rather than only transformed ones, so behaviour-neutral on the existing suite is the intended result.

andymai:
Correct again, and the cause is an adapter disagreement I should have checked.

`interpolatePoints` is not consistent across the two OCCT adapters:

- `src/kernel/occtWasm/curveOps.ts:95` forwards `periodic` to the kernel.
- `src/kernel/occt/curveOps.ts:34` destructures only `tolerance` and builds an open B-spline through `GeomAPI_PointsToBSpline_2`.

`resolveOcct()` prefers `'occt'`, so the adapter that silently drops the flag is the one replay actually runs against. Removing the duplicate endpoint and asking for `periodic` left the seam open by the width of the final segment — measured 0.1227 on a 24-sample circle.

Fixed in 1c290e2a by keeping the repeated final point and interpolating through it, which closes the curve geometrically on both adapters rather than depending on a flag only one honours. Weaker than parametric periodicity, but correct regardless of which kernel resolves.

The new test asserts the seam directly rather than trusting the option: it evaluates the curve at both parameter ends and requires the gap under 1e-6, and checks the bounding box is the sheared ellipse (4R x 2R) rather than the source circle. Verified it fails at 0.1227 against the previous version.

Manifold suite: 620 failures before and after, no regressions.

PR #2222, opened 2026-08-22
feat(families): route BIM export on a declared archetype, not the family name

andymai:
## The problem

`familiesAdapter.ts` keyed the entire IFC mapping on the family's **display name**. `SPEC_ROUTES` was `{ Wall, Slab, Column, Beam, Roof, Stair }`, the spatial container was a literal `el.type === 'Storey'`, and `ResolvedElement.type` is just `familyName`. So every starter-registry family name was an undeclared API contract with a different package.

That sits badly with copy-in distribution. `registry/families/storey.ts` carries `// brepjs-family: storey@1` and is explicitly handed over for the user to own, and renaming a file you own is the obvious thing to do, especially with `Storey` (the IFC spelling, not the one most teams use).

The two failure modes:

- **Renamed `Storey`** → `FAMILIES_NO_STOREY`, naming the thing you just renamed away.
- **Renamed `Wall`** → falls out of `SPEC_ROUTES`. Without `proxyEvaluator` that is a hard `FAMILIES_UNSUPPORTED_TYPE`; with one set it becomes an `IfcBuildingElementProxy`, so you get a valid IFC file full of proxies where walls should be, and no signal.

## The change

Families declare an `archetype`, and the adapter routes on that:

```ts
family('Storey', render, { archetype: 'storey' })
family('Level',  render, { archetype: 'storey' })  // same mapping, different name
```

`archetype` is an opaque string that brepjs-families never interprets. The vocabulary belongs to the adapter, which is what keeps the family layer free of IFC (`src/index.ts`: "Domain-neutral: no BIM imports here"). It sits beside `role: 'fill'`, which already decoupled "fills an opening" from the family's name; this extends the same idea to the routed element types and the spatial container.

`SPEC_ROUTES` is rekeyed to archetypes, and every name check on the routing path goes through `archetypeFor(el)`: the route lookup, the storey container, the wall-opening gate, the synthesized-`Opening` suppression in the child walk, and `addFill`'s `IfcDoor`-vs-`IfcWindow` choice. Mapped archetypes are `storey`, `wall`, `slab`, `column`, `beam`, `roof`, `stair`, `door`, `window`.

The two opening-path checks are the ones that have to agree. Greptile caught them disagreeing in the first commit: `addOpenings` was still gated on `el.type === 'Wall'` while the child walk already suppressed the `Opening` on `archetype === 'wall'`, so a renamed wall skipped the mapping *and* swallowed the child, and its door left no trace in an otherwise valid file. Following that up found the same defect in `addFill`, which picked the IFC entity off `fill.type`. `role: 'fill'` makes opening *synthesis* name-independent, which is why fillers looked like they needed no archetype; the entity choice was still name-keyed.

## Compatibility

Undeclared families fall back to matching their display name, so this is purely additive. **All 659 existing brepjs-bim tests pass unchanged.**

The fallback keeps the original trap alive for undeclared families, which is why two other things changed:

- Both error messages now name archetype as the fix, and `FAMILIES_UNSUPPORTED_TYPE` reports the archetype it resolved (`none` when neither a declaration nor the name fallback matched).
- `FamiliesBimResult` gains `proxied: readonly ProxiedElement[]`, so a lost route shows up in the result instead of only in the file. It is only ever non-empty when `proxyEvaluator` is set, since without it an unrouted element is still a hard error.

## Also in here

- Starter registry families declare their archetypes; version markers and manifest entries bump (`storey@2`, `wall@2`, `slab@2`, `column@3`, `beam@2`, `roof@3`, `stair@2`, `door@2`, `window@2`).
- Docs: a routing section in `families/ifc-export`, a renaming section in `families/copy-in`, the two error entries in `reference/errors`, a glossary entry, the brepjs-cad `families-bim` skill reference, and the families README.
- Playground ambient types regenerated.
- Prettier reflowed two pre-existing lines in `registry/families/wall.ts`; that file was already unformatted on `main` (`registry/` is outside the `format:check` globs).

## Verification

- brepjs-bim 664 tests (659 existing + 5 new), brepjs-families 59 (55 + 4)
- New tests: renamed `Level`/`Partition` still reach `IFCBUILDINGSTOREY`/`IFCWALL`; an undeclared rename is reported in `proxied`; an archetype beats a display name that would route elsewhere (family named `Wall`, `archetype: 'column'`, lands as `IfcColumn`); a renamed wall keeps its opening, filler and both relationships; a renamed window filler still reaches `IFCWINDOW`
- The two opening tests were verified RED against the previous commit and GREEN here, and only those two moved
- typecheck (root, families incl. jsx config, bim, playground), eslint, knip, boundaries, check:patterns all clean

### DIFF
--- a/apps/docs/families/copy-in.md
+++ b/apps/docs/families/copy-in.md
@@ -56,6 +56,16 @@ npx brepjs diff wall
 
 `diff` compares your copy against the registry: stale version markers are called out (`local wall@1, registry wall@2`), content drift renders as a git diff, and the exit code is 1 on any difference, which makes it usable as a CI guard for teams that want to know when their copies diverge from upstream.
 
+## Renaming what you copied
+
+A copied family is yours, including its name, and `Storey` in particular is the IFC spelling rather than the one every team uses. Rename it freely; the `archetype` in the file is what a BIM export routes on, not the name:
+
+```typescript
+export const Level = family('Level', render, { archetype: 'storey', props: storeySchema });
+```
+
+Keep the `// brepjs-family: storey@2` marker on the first line whatever you call the family. `diff` anchors on it, and it is how you find out the upstream file moved on.
+
 ## Self-hosting a registry
 
 The registry is data: a `manifest.json` plus source files, no server logic.
--- a/apps/docs/families/ifc-export.md
+++ b/apps/docs/families/ifc-export.md
@@ -60,6 +60,8 @@ family('Level', render, { archetype: 'storey' }); // still IfcBuildingStorey
 family('Partition', render, { archetype: 'wall' }); // still IfcWall
 ```
 
+Fillers route the same way: `archetype: 'door'` or `'window'` on a `role: 'fill'` family picks the IFC entity, so a `Glazing` family still becomes an `IfcWindow`. The mapped archetypes are `storey`, `wall`, `slab`, `column`, `beam`, `roof`, `stair`, `door`, and `window`.
+
 `archetype` is a plain string that brepjs-families never interprets: the vocabulary belongs to the adapter, which is what keeps the family layer free of IFC.
 
 A family that declares none falls back to matching its display name, so families written before `archetype` existed keep working. The fallback carries the original trap, though: rename an undeclared family and it quietly stops routing. Without `proxyEvaluator` that is a hard `FAMILIES_UNSUPPORTED_TYPE` error, but with one set it becomes an `IfcBuildingElementProxy` instead of the element you meant, which is a valid file that is wrong. Every proxied element is reported so you can catch it:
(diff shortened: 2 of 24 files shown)


andymai:
Good catch, fixed in cc7b74b9. The wall case was worse than a missed conversion: `addOpenings` was gated on `el.type === 'Wall'` while the child walk suppressed the synthesized `Opening` on `archetype === 'wall'`, so a renamed wall skipped the mapping *and* swallowed the child. The door left no trace at all in an otherwise valid file.

Following the thread found the same defect one layer down, which the review didn't flag: `addFill` picked `IfcDoor` vs `IfcWindow` off `fill.type`, so a renamed filler hit `FAMILIES_UNSUPPORTED_FILL`. `role: 'fill'` makes opening *synthesis* name-independent, which is what led me to skip archetypes on fillers in the first place, but the IFC entity choice was still name-keyed. `door` and `window` are now part of the archetype vocabulary and the starter fillers declare them.

Two regression tests added, both verified to fail against b8ae9d83 and pass here: a renamed wall keeps its opening, filler and both relationships through to IFC, and a renamed window filler still reaches `IFCWINDOW`.

— claude

PR #2257, opened 2026-08-27
fix(viewer): use brepjs Z-up coordinates in previews

viktar-b:
## What does this PR do?

Fixes incorrect axis behavior in `brep preview`.

brepjs geometry uses CAD coordinates where X/Y form the ground plane and +Z is up. The shared viewer was configured around Y-up coordinates, causing its camera presets, orbit controls, and grid to disagree with the model coordinate system.

This PR:

- Configures perspective and orthographic cameras with +Z as up.
- Updates the Iso, Front, Top, and Right camera directions.
- Places the grid on the XY plane at `z = -0.01`.
- Converts scene lighting to Z-up coordinates.
- Removes the playground's Z-up-to-Y-up geometry rotation.
- Updates playground bounds, camera fitting, and presets to use native model coordinates.
- Documents the viewer coordinate convention.
- Adds regression tests for camera and grid orientation.

## How to test

Run the automated checks:

```bash
npm -w packages/brepjs-viewer test
npm -w packages/brepjs-viewer run typecheck
npm -w packages/brepjs-viewer run lint
npm -w packages/brepjs-viewer run build
npm -w apps/playground run typecheck
npm -w packages/brepjs-cad test -- --run tests/preview.test.ts
npm -w packages/create-brepjs test -- --run tests/scaffold.test.ts
```

For manual verification, install the locally packed `brepjs-cad` into a fresh BIM scaffold and run:

```bash
npm run preview
```

Confirm that the slab lies on the XY ground plane, the wall rises along +Z, and the Iso, Front, Top, and Right presets use the expected model-space axes.

## Checklist

- [x] Tests pass (`npm run test`)
- [x] Lint passes (`npm run lint`)
- [x] Layer boundaries checked (`npm run check:boundaries`)
- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)


andymai (approved):
Reviewed. This is a clean, mathematically consistent Y-up to Z-up conversion. Approving.

## Correctness

Every axis vector in the PR is the exact inverse of the removed geometry rotation. The old playground applied a -90 degree rotation about X to convert model Z-up into three.js Y-up, which maps (x, y, z) to (x, z, -y). Removing that rotation and expressing all camera/light/grid vectors directly in model space requires applying the inverse (x, y, z) to (x, -z, y). I checked each one against that inverse:

- Camera presets (useCameraPresets and ViewerCanvas VIEW_DIR): front, side/right, top, iso all convert exactly.
- Directional lights: [-50, 60, 80] to [-50, -80, 60] and [40, -40, 30] to [40, -30, -40], both exact.
- Default camera positions: playground [40, 30, 40] to [40, -40, 30], exact.
- Grid: plane un-rotated to native XY, position moved from y=-0.01 to z=-0.01, shader sample changed xz to xy. Consistent.
- hemisphereLight sky direction set to +Z. Correct for Z-up.

The `top` preset keeps a small off-axis nudge (0, 0.01, 1) against up=[0,0,1], which avoids a degenerate camera basis (cross product is non-zero). Good.

## Tests

The two new regression tests lock in the grid orientation and the Z-up framing/orbit-up, which are exactly the things a future refactor could silently break. Solid coverage.

## CI note

The red `ci-pass` here is not a code failure. The `benchmark` job ran both benchmark passes successfully and then failed only at its final step, which posts a results comment via the Actions token. Fork PRs get a read-only `GITHUB_TOKEN`, so that POST returns 403 "Resource not accessible by integration". The benchmark job is informational only (it never fails on an actual regression), so this is purely a fork-permissions artifact. Merging past it. I am opening a separate PR to guard that comment step so future external contributions do not trip the required gate.

